PR #19, opened 2019-11-19
BUG 1774078: fix env var name

zshi-redhat:


hack/env.sh
```diff
@@ -6,4 +6,4 @@ export RELEASE_VERSION=v4.3.0
 export OPERATOR_NAME=ptp-operator
 
 LINUXPTP_DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-ptp | jq --raw-output '.Digest')
-export LINUXPTP_DAEMON_IMAGE=${LINUXPTP_DAEMON_IMAGE:-quay.io/openshift/origin-ptp@${LINUX_DAEMON_IMAGE_DIGEST}}
```
e-minguez: Actually, it doesn't work to me unless I use `:+` assignment:

```
export LINUXPTP_DAEMON_IMAGE=${LINUXPTP_DAEMON_IMAGE:+quay.io/openshift/origin-ptp@${LINUXPTP_DAEMON_IMAGE_DIGEST}}
```

e-minguez: Nevermind, it works with the proposed fix. PEBKAC.

PR #26, opened 2019-12-05
BUG 1775630: fix olm.skipRange version

zshi-redhat:


manifests/4.3/ptp-operator.v4.3.0.clusterserviceversion.yaml
```diff
@@ -12,7 +12,7 @@ metadata:
     createdAt: 2019-10-14
     certified: "false"
     repository: https://github.com/openshift/ptp-operator
-    olm.skipRange: ">=4.3.0 <4.4.0"
+    olm.skipRange: ">=4.3.0 <4.3.0"
```
s1061123: Does it means skip everything (nothing more than+equal 4.3.0 and less than 4.3.0)?

zshi-redhat: Yes, it means if the version is in that range, then don't auto upgrade the operator.
In above case, it should skip 4.3.0 version and upgrade to any version that larger than 4.3.0, for example, if there is a z-stream version 4.3.1.

PR #40, opened 2020-03-04
BUG 1810129: [4.4] match regex rule defined in art.yaml

zshi-redhat:


zshi-redhat:
close PR, the is created againt wrong branch.

PR #46, opened 2020-04-13
Fix cluster-scope issue

SchSeba:
This commit also change the env.sh file to use latest image.

The scope is related to the controller-runtime update.

https://github.com/kubernetes-sigs/controller-runtime/pull/675/files

### DIFF
--- a/hack/env.sh
+++ b/hack/env.sh
@@ -2,7 +2,8 @@ REPO_DIR="$(dirname $0)/.."
 NAMESPACE=openshift-ptp
 OPERATOR_EXEC=oc
 
-export RELEASE_VERSION=v4.3.0
+export RELEASE_VERSION=v4.5.0
+export IMAGE_TAG=latest
 export OPERATOR_NAME=ptp-operator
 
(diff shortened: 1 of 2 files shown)


hack/env.sh
```diff
@@ -2,10 +2,9 @@ REPO_DIR="$(dirname $0)/.."
 NAMESPACE=openshift-ptp
 OPERATOR_EXEC=oc
 
-export RELEASE_VERSION=v4.3.0
+export RELEASE_VERSION=v4.5.0
+export IMAGE_TAG=latest
```
pliurh, line 6: The reason to use the digest rather than the `latest` is that, we want the host to always pull the latest image, instead of using the cached one with the `latest` label.

SchSeba: @pliurh thanks for the comment I revert the change to use the digest again.
I leave this IMAGE_TAG variable here because when we cut a branch we must point to the release version on that branch.

I open other PR's here to fix that issue.
when we deploy ptp-operator for the CI it was deploying the latest one and not the 4.3 images.

zshi-redhat: Makes sense, I think we can use image tag for released branches and use digest for master.

zshi-redhat:
current change looks good to me.

PR #48, opened 2020-04-13
Fix env file for the operator test deployment

SchSeba:


zshi-redhat:
@SchSeba do we still need this PR for CI? it requires a bz be associated.

SchSeba:
let's close this one as it required a BZ and this only affects the CI.
If there is a request for this I will reopen it and create a Bugzilla for it

/close

PR #50, opened 2020-04-15
Bug 1826302: Make the tests exportable.

fedepaol:
Have them belong to a different package, so they can be vendored.
Do not panic if the client does not get created, but return nil and check for it not to be nil instead.
This allows us to dry run and check what tests are available.


fedepaol:
/hold
needs to be validated.

PR #51, opened 2020-04-19
Added new ptp test which validates if metric is present on slave

kononovn:


SchSeba:
@kononovn I was thinking on something like this
```
Describe("PTP Metric", func() {
		ptpRunningPods := []v1core.Pod{}
		BeforeEach(func() {
			ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
			Expect(err).NotTo(HaveOccurred())
			Expect(len(ptpPods.Items)).To(BeNumerically(">", 0), fmt.Sprint("linuxptp-daemon is not deployed on cluster"))
			for _, pod := range ptpPods.Items {
				if podRole(pod, PtpSlaveNodeLabel) || podRole(pod, PtpGrandmasterNodeLabel) {
					waitUntilLogIsDetected(pod, 3*time.Minute, "Profile Name:")
					ptpRunningPods = append(ptpRunningPods, pod)
				}
			}
			Expect(len(ptpRunningPods)).To(BeNumerically(">=", 2), fmt.Sprint("Fail to detect PTP slave/master pods on Cluster"))

		})

		Context("is present", func() {

			// 27324
			It("on slave", func() {
				var timeDiff string
				slavePodDetected := false
				for _, pod := range ptpRunningPods {
					if podRole(pod, PtpSlaveNodeLabel) {
						Eventually(func() string {
							buf, _ := pods.ExecCommand(client.Client, pod, []string{"curl", "127.0.0.1:9091/metrics"})
							timeDiff = buf.String()
							return timeDiff
						}, 3*time.Minute, 2*time.Second).Should(ContainSubstring("openshift_ptp_max_offset_from_master"),
							fmt.Sprint("Time metrics are not detected"))
						slavePodDetected = true
					}
				}
				Expect(slavePodDetected).ToNot(BeFalse(), "No slave pods detected")
			})
		})
	})
```

I test it and it pass

```
• [SLOW TEST:38.766 seconds]
ptp
/home/sscheink/Documents/GolangProjects/src/github.com/openshift/ptp-operator/test/ptp.go:29
  PTP Metric
  /home/sscheink/Documents/GolangProjects/src/github.com/openshift/ptp-operator/test/ptp.go:304
    is present
    /home/sscheink/Documents/GolangProjects/src/github.com/openshift/ptp-operator/test/ptp.go:320
      on slave
      /home/sscheink/Documents/GolangProjects/src/github.com/openshift/ptp-operator/test/ptp.go:323
------------------------------

Ran 1 of 9 Specs in 51.832 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 8 Skipped
PASS | FOCUSED

Debugger finished with exit code 0
```

SchSeba:
@kononovn can you please rebase this PR?

test/ptp/ptp.go
```diff
@@ -304,6 +304,44 @@ var _ = Describe("[ptp]", func() {
...
+
+			// 27324
+			It("on slave", func() {
+				var timeDiff string
+				slavePodDetected := false
+				for _, pod := range ptpRunningPods {
+					if podRole(pod, PtpSlaveNodeLabel) {
+						Eventually(func() string {
+							buf, _ := pods.ExecCommand(client.Client, pod, []string{"curl", "127.0.0.1:9091/metrics"})
+							timeDiff = buf.String()
```
fedepaol, line 334: Just return buf.String and get rid of the `timeDiff` variable.

test/ptp/ptp.go
```diff
@@ -304,6 +304,44 @@ var _ = Describe("[ptp]", func() {
 			waitUntilLogIsDetected(testPtpPod, 3*time.Minute, "Profile Name: slave")
 		})
 	})
+
+	Describe("PTP Metric", func() {
+		ptpRunningPods := []v1core.Pod{}
+		BeforeEach(func() {
+			ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
```
fedepaol, line 311: I would get rid of this and put the metric test under PTP e2e tests.
If we don't do that, we'll have to replicate all the logic there to check / configure ptp / discovery mode.
Here we are checking that the daemons are running, but we may even need to configure those before. The test will fail if we focus only on PTP metric because the setup won't happen.
Also, we rely on the test order execution assuming the setup was done before.

PR #53, opened 2020-04-23
Bug 1828692: Provide support for running tests in a disconnected cluster.

fedepaol:


fedepaol:
/hold
Let's see if we get official images on quay

SchSeba:
I think we need to remove this 

https://github.com/openshift/ptp-operator/blob/92f3d20c48f9b8e75e78bc2579c01937c44ef830/test/utils/pods/pods.go#L22

and also check if we use this image 
https://github.com/openshift/ptp-operator/blob/92f3d20c48f9b8e75e78bc2579c01937c44ef830/test/utils/pods/pods.go#L37

fedepaol:
> hugePageCommand

Removed both

fedepaol:
The image is not even used, we will clean up the test utils later.

PR #55, opened 2020-05-08
Bug 1830496: Remove runlevel from the operator's namespace and use SCC

fedepaol:
Use the privileged scc for the ptp-daemon.


zshi-redhat:
@fedepaol please also add role.yaml and rolebinding.yaml in `hack/undeploy.sh`

fedepaol:
> @fedepaol please also add role.yaml and rolebinding.yaml in `hack/undeploy.sh`

Ah right, I missed that.

PR #58, opened 2020-06-16
Provide an optional node selector to choose the nodes impacted by the tests

fedepaol:
While running tests on a cluster, we may want to limit the number of nodes impacted by those nodes.
By providing an optional selector we choose the nodes where to apply the configuration.

Also, refactor the cleaning logic in a separate function so we can invoke it.

### DIFF
--- a/test/ptp/ptp.go
+++ b/test/ptp/ptp.go
@@ -297,7 +297,7 @@ func configurePTP() {
 	err := clean.All()
 	Expect(err).ToNot(HaveOccurred())
 
-	ptpNodes, err := nodes.GetNodeTopology(client.Client)
+	ptpNodes, err := nodes.PtpEnabled(client.Client)
 	Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/ptp/ptp.go
```diff
@@ -262,33 +263,10 @@ var _ = Describe("[ptp]", func() {
...
-	}
-
-	nodeList, err := client.Client.Nodes().List(metav1.ListOptions{LabelSelector: fmt.Sprintf("%s=", PtpGrandmasterNodeLabel)})
-	Expect(err).ToNot(HaveOccurred())
-	for _, node := range nodeList.Items {
-		delete(node.Labels, PtpGrandmasterNodeLabel)
-		_, err = client.Client.Nodes().Update(&node)
-		Expect(err).ToNot(HaveOccurred())
-	}
+	clean.All()
```
SchSeba, line 266: you need to get the error here and check it

PR #59, opened 2020-06-16
add 4.6 Operator manifests

zshi-redhat:


manifests/4.6/ptp-operator.v4.5.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,244 @@
+apiVersion: operators.coreos.com/v1alpha1
```
fedepaol, line 1: Should the filename be v4.6.0.xxxx ?

zshi-redhat: Yes, it should be. I forgot to change it when copying from 4.5.  Thanks @fedepaol !
Updated.

PR #71, opened 2020-09-26
Bug 1880990: Choose the interfaces between master and slave.

fedepaol:
In order to make sure ptp tests are reliable, we look for interfaces compatible between slave and master before running the
tests.


test/ptp/ptp.go
```diff
@@ -305,6 +281,7 @@ var _ = Describe("[ptp]", func() {
 				slavePodDetected := false
 				for _, pod := range ptpRunningPods {
 					if podRole(pod, slaveNodeLabel) {
+						fmt.Println("Checking metric on ", pod.Name)
```
mmirecki, line 284: Leftover debug message

PR #73, opened 2020-10-02
Bug 1884529: Skip the higher profile priority test in discovery mode

fedepaol:
In discovery mode we don't want to change the current configuration.


SchSeba (approved):
/lgtm

PR #74, opened 2020-10-02
Bug 1884547: Test PTP daemon apply match rule based on nodeLabel: use the discovered profile.

fedepaol:
Right now it uses the hardcoded profile, which is fine for regular mode, but not good for discovery mode.

### DIFF
--- a/test/ptp/ptp.go
+++ b/test/ptp/ptp.go
@@ -90,6 +90,7 @@ var _ = Describe("[ptp]", func() {
 		var ptpRunningPods []v1core.Pod
 		var masterNodeLabel, slaveNodeLabel string
 		discoveryFailed := false
+		var masterProfile, slaveProfile string
 
 		execute.BeforeAll(func() {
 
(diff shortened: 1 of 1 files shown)


fedepaol:
/hold
still validating

test/ptp/ptp.go
```diff
@@ -171,16 +176,17 @@ var _ = Describe("[ptp]", func() {
 
 			// 25733
 			It("PTP daemon apply match rule based on nodeLabel", func() {
-				profileSlave := "Profile Name: test-slave"
-				profileMaster := "Profile Name: test-grandmaster"
+				profileSlave := fmt.Sprintf("Profile Name: %s", slaveProfile)
+				profileMaster := fmt.Sprintf("Profile Name: %s", masterProfile)
```
SchSeba, line 180: nit: can you print the master profile only if it's not empty string?

SchSeba (approved):
/lgtm

PR #75, opened 2020-10-07
Update paths so that copy works, update to use go 1.15 images

dougbtv:
Got a heads up from ART about rhel-8 and golang-1.15 not being optional, went to run locally, and found the path issue.

### DIFF
--- a/must-gather/Dockerfile
+++ b/must-gather/Dockerfile
@@ -1,8 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
-WORKDIR /go/src/github.com/openshift/ptp-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/ptp-operator/must-gather
 COPY . .
 
-FROM quay.io/openshift/origin-must-gather:4.3.0
+FROM registry.svc.ci.openshift.org/ocp/4.7:must-gather
 COPY --from=builder /go/src/github.com/openshift/ptp-operator/must-gather/collection-scripts/* /usr/bin/
+RUN chmod +x /usr/bin/gather
 
 ENTRYPOINT /usr/bin/gather
--- a/must-gather/Dockerfile.rhel7
+++ b/must-gather/Dockerfile.rhel7
@@ -1,10 +1,11 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
-WORKDIR /go/src/github.com/openshift/ptp-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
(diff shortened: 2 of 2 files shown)


fedepaol:
Thanks!
/lgtm

must-gather/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
-WORKDIR /go/src/github.com/openshift/ptp-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+WORKDIR /go/src/github.com/openshift/ptp-operator/must-gather
 COPY . .
 
-FROM quay.io/openshift/origin-must-gather:4.3.0
+FROM registry.svc.ci.openshift.org/ocp/4.7:must-gather
```
yselkowitz, line 5: What is the reason for two different Dockerfiles in this directory?  If this one is meant to be for external building by hand for developers, and not used in CI or ART builds, then it can stay as it was (probably with a later version).

must-gather/Dockerfile.rhel7
```diff
@@ -1,8 +1,8 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
-WORKDIR /go/src/github.com/openshift/ptp-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-7-golang-1.15-openshift-4.6 AS builder
```
yselkowitz, line 1: As above, what is the reason for two different Dockerfiles in this directory?  If this one is meant to be used by CI or ART builds, then the "rhel7" shouldn't be taken literally anymore, and rhel-8-golang-1.15-openshift-4.6 should be used instead.

yselkowitz (changes requested):

must-gather/Dockerfile
zshi-redhat: @yselkowitz you got it right, the Dockerfile w/o suffix is used for external building.
The one w suffix Dockerfile.rhel7 is used for ART building.

must-gather/Dockerfile.rhel7
zshi-redhat: Yes, we shall use rhel-8.

zshi-redhat (commented):
@dougbtv thanks for the fix!
added a few comments. mostly to clarify the two dockerfile usages.

must-gather/Dockerfile.rhel7
```diff
@@ -1,10 +1,11 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
-WORKDIR /go/src/github.com/openshift/ptp-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
```
dougbtv, line 1: Updated to RHEL 8

yselkowitz:
As long as you're satisfied with the external buildability with the plain (non-CI/ART) Dockerfile, then:
/lgtm

PR #82, opened 2020-11-18
Bug 1898531: Tests: Adjust ptp configuration in case of ipv6 only interfaces.

fedepaol:
When checking for available interfaces, we use the default for ptp4l, which is ipv4.
Here, we check if a given interface has ipv6 only addresses, and in case we add the -6 parameter
to ptp4l.

### DIFF
--- a/test/ptp/ptp.go
+++ b/test/ptp/ptp.go
@@ -330,7 +330,7 @@ func configurePTP() {
 			By("Creating the policy for the grandmaster node")
 			err = createConfig(PtpGrandMasterPolicyName,
 				gmInterface,
-				"",
+				"-2",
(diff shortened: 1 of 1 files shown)


fedepaol:
/hold
need to validate against an ipv4 cluster

kenyis:
Just wanted to follow up on our discussion this morning.  We definitely want to test Layer 2 most of the time.  Therefore, whether the installation is IPv4 or IPv6, we should be using Layer 2.  Therefore, the "-2" flag is set regardless.

fedepaol:
@SchSeba ptal

SchSeba (approved):
/lgtm

PR #84, opened 2020-12-14
Fixed discovery mode ptp tests for SNO.

kononovn:
Allow to run tests in discovery mode with 1 ptp pod running.

### DIFF
--- a/test/ptp/ptp.go
+++ b/test/ptp/ptp.go
@@ -146,17 +146,29 @@ var _ = Describe("[ptp]", func() {
 				if discoveryFailed {
 					Skip("Failed to find a valid ptp slave configuration")
 				}
-				ptpRunningPods = []v1core.Pod{}
 				ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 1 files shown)


test/ptp/ptp.go
```diff
@@ -156,7 +156,11 @@ var _ = Describe("[ptp]", func() {
 						ptpRunningPods = append(ptpRunningPods, pod)
 					}
 				}
-				Expect(len(ptpRunningPods)).To(BeNumerically(">=", 2), fmt.Sprint("Fail to detect PTP slave/master pods on Cluster"))
+				expectedPtpRunningPods := 2
+				if !discovery.Enabled() {
+					expectedPtpRunningPods = 1
```
fedepaol, line 161: Shouldn't be the other way around? 1 pod in discovery mode (only slave).
Also, I think it would make sense to check we found a slave splitting the or above.

test/ptp/ptp.go
```diff
@@ -303,7 +307,11 @@ var _ = Describe("[ptp]", func() {
 						ptpRunningPods = append(ptpRunningPods, pod)
 					}
 				}
-				Expect(len(ptpRunningPods)).To(BeNumerically(">=", 2), fmt.Sprint("Fail to detect PTP slave/master pods on Cluster"))
+				expectedPtpRunningPods := 2
+				if !discovery.Enabled() {
```
fedepaol, line 311: Same

test/ptp/ptp.go
kononovn: Oh yes you are right. Changed. Please review it once again

PR #87, opened 2021-01-10
Fix the issue:"Master will not be selected if there is a configution - ptp4lOpts: '--summary_interval'"

evgenLevin:


SchSeba (commented):
/ok-to-test

PR #91, opened 2021-01-13
Update ptp-daemon.yaml to expose the UDS socket

williamcaban:
Updates to bindata/ptp-daemon.yaml to expose the UDS socket from /var/run inside the container into host path /var/run/ptp so it can be access by applications running in the node requiring to interact with ptp4l for tracking PTP time synchronization

### DIFF
--- a/bindata/linuxptp/ptp-daemon.yaml
+++ b/bindata/linuxptp/ptp-daemon.yaml
@@ -66,6 +66,8 @@ spec:
         volumeMounts:
         - name: config-volume
           mountPath: /etc/linuxptp
+        - name: socket-dir
+          mountPath: /var/run
       volumes:
         - name: config-volume
           configMap:
@@ -73,6 +75,10 @@ spec:
         - name: linuxptp-certs
           secret:
             secretName: linuxptp-daemon-secret
+        - name: socket-dir
+          hostPath:
+            path: /var/run/ptp
+            type: DirectoryOrCreate
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor


SchSeba:
General comment:

from the man page https://linux.die.net/man/8/pmc

if we expose the uds socket in the host every pod will be able to request the host mount and send `SET` commands that can change/break the time synchronization on the node.

This is my only concern about this PR

@zshi-redhat @williamcaban WDYT?

williamcaban:
> General comment:
> 
> from the man page https://linux.die.net/man/8/pmc
> 
> if we expose the uds socket in the host every pod will be able to request the host mount and send `SET` commands that can change/break the time synchronization on the node.
> 
> This is my only concern about this PR
> 
> @zshi-redhat @williamcaban WDYT?

Correct. That is how a ptp4l UDS socket is expected to work. When PTP is enabled on a host is because the workload running on it require that level of timing accuracy and, at least on Telco, that includes direct access to the PTP UDS socket for specific interactions controlled/managed by the RAN vendor.

We can agree that from K8s perspective this is NOT a multi-tenant aware approach (just like when ptp4l is running on a Linux host). Due to the lack of existence of a more granular specifications or standards for this purpose, applications consuming the PTP service using or based on PMC, require access to the UDS socket.  If we want or need to provide read-only or authenticated options for the UDS socket, that should be work on the linuxptp project. Similar to how the access to a /var/run/crio/crio.sock is defined in the CRIO project.

On the other hand, from Telco perspective, when a workload like a RAN DU is on a node, it expects to be able to interact with the PTP UDS socket. I'm not aware of a Telco use case for RAN where PTP is needed and access to the ptp4l socket it is not required.

Brainstormed options and why to avoid them:
- Exposing read-only hostPath: due to the way the ptp4l UDS socket works, the client must be able to write to the socket to retrieve information
- As mitigation use an actual PVC to expose the UDS: Since there is a 1:1 ratio of PVs to PVCs and PVCs are namespaces, this approach prevents applications requiring access to the UDS, like Telco DU workloads which run in their own namespace, from accessing the PVC without deploying some sort of helper application in the same namespace as the ptp operator which in turn will require additional privileges. 

Long term option (out of scope):
- Define an open API's to expose the synchronization for the workloads with standard interfaces: At this moment, due to the lack of existence of such specifications or standard for this purpose, and the fact the requirements are specific to the CNF workload type running in the node, this is a long term option that need to be worked with open specification bodies and are out of scope for this operator.

PTP is a protocol used on specialized environment for specialized workload, as such I propose:
- Lets document the Operator does not add any additional security or control mechanism over the UDS to avoid any confusion
- (even when I don't see a use case for PTP without the UDS) - we can extend the Operator API to allow enabling/disabling exposing the UDS to the host in case there is a use case for using PTP on a node without requiring access to the UDS socket.

@SchSeba @zshi-redhat - does it make sense to also start working with linuxptp on enhancing it to provide more granular access control for the UDS socket?

beekhof:
What if we make this change optional, so that non-telco installations are unaffected?

beekhof:
Actually, talking to @zshi-redhat this is an optional operator delivered via OLM specifically for telco.
In which case, it doesn't make sense to make this change conditional after-all, and I would strongly advocate for merging the PR as-is for 4.8.

beekhof:
> does it make sense to also start working with linuxptp on enhancing it to provide more granular access control for the UDS socket?

IMHO, the long term solution we should be focusing on is about building the general notification solution that makes access to the socket irrelevant.

SchSeba:
I validate this PR and it's working only if the pod that want to use the UDS socket is a privileged pod so I am fine merging this PR.

I will open a follow PR with a functional test for this

PR #93, opened 2021-02-09
Bump protobuf to v1.3.2

SchSeba:
Update gogo/protobuf to v1.3.2 to fix https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3121. The related k8s fix is here: kubernetes/kubernetes#98477


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32 // indirect
 	github.com/go-openapi/spec v0.19.4
 	github.com/gofrs/uuid v3.2.0+incompatible // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
```
fedepaol, line 19: To be sure we consume ONLY 1.3.2 you need to pin it with replace, otherwise indirect dependencies will keep using the previous version (see the go.sum file)

SchSeba: thanks! can you have another look?

zshi-redhat:
/hold
for Federico to take a second look.

PR #96, opened 2021-02-11
Skip tests in single node enviroments

mmirecki:


test/ptp/ptp.go
```diff
@@ -30,6 +30,14 @@ import (
 	"github.com/openshift/ptp-operator/test/utils/pods"
 )
 
+var isSingleNode bool
+
+func init() {
+	var err error
+	isSingleNode, err = nodes.IsSingleNode()
```
fedepaol, line 37: Move the call inside the tests (beforeAll for example)

test/ptp/ptp.go
```diff
@@ -102,7 +110,9 @@ var _ = Describe("[ptp]", func() {
 		var masterProfile, slaveProfile string
 
 		execute.BeforeAll(func() {
-
+			if isSingleNode {
+				Skip("Running in single node mode")
```
fedepaol, line 114: Can you expand the reason of the skipping?
Also, this won't work because beforeAll is a beforeEach that runs only the first time.
The result will be that only the first test will be skipped.
Add a beforeEach that does the skipping

test/utils/nodes/nodes.go
```diff
@@ -101,3 +101,11 @@ func MatchingOptionalSelectorPTP(toFilter []ptpv1.NodePtpDevice) ([]ptpv1.NodePt
 	}
 	return res, nil
 }
+
+func IsSingleNode() (bool, error) {
```
fedepaol, line 105: Either move to the cluster package or rename it

fedepaol:
@mmirecki can you expand the commit message with the reason for skipping?

test/ptp/ptp.go
mmirecki: Done

test/utils/nodes/nodes.go
mmirecki: Renamed

test/ptp/ptp.go
mmirecki: This is actually meant to skip the configuration in beforeAll (which would fail for sno).

SchSeba:
@mmirecki this looks good there is a reason for the hold?

test/ptp/ptp.go
fedepaol: Yes ,but if you skip in beforeAll, it's gonna skip only the first test of the suite, not all of them

mmirecki: There is also a skip in beforeEach.
Not sure we can have the Skip directly in Describe

fedepaol: Ah, sorry I did not notice it.
Thanks

PR #98, opened 2021-02-25
Fix mustgather 4.7

oribon:


SchSeba:
there is no need for this PR any more the ART team handle the problem

PR #104, opened 2021-03-30
Reduce CPU request for the linuxptp-daemonset

SchSeba:
This commit reduce the amount of cpu requested by the linuxptp-daemon


SchSeba:
/cc @zshi-redhat 

We need this PR for SNO

PR #105, opened 2021-04-08
Update operator-sdk to 1.5.0

oribon:


zshi-redhat:
@oribon I assume this would be a 4.9 change, right?

oribon:
> @oribon I assume this would be a 4.9 change, right?

@zshi-redhat  Yes, AFAIK

controllers/suite_test.go
```diff
@@ -0,0 +1,83 @@
+/*
```
mmirecki, line 1: No tests in this suite

PR #107, opened 2021-04-11
Move ptp-daemon files to this repository 

oribon:
depends on https://github.com/openshift/ptp-operator/pull/105


cmd/linuxptp-daemon/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM fedora:32 AS builder
```
mmirecki, line 1: We now have some Dockerfiles cmd/.., and some in root dir
Maybe ./Dockerfile.ptp-daemon  ?

cmd/linuxptp-daemon/README.md
```diff
@@ -0,0 +1,126 @@
+# Linuxptp Daemon for Kubernetes
```
mmirecki, line 1: We should also add a mention in the main readme about having the daemon.

PR #108, opened 2021-04-18
Revert "Update operator-sdk to 1.5.0"

SchSeba:
Reverts openshift/ptp-operator#105


zshi-redhat:
@SchSeba I understand that we accidentally merged this in 4.8, which shall be done in 4.9, thus the revert, right?

SchSeba:
@zshi-redhat thanks for the approval and yes this should be part of 4.9 not 4.8.

I need to check with the CI team that PR didn't have the `approve` and it also contain the `needs-ok-to-test` so I have not idea why tide merge it

PR #110, opened 2021-05-12
PTP Tests: check ethtool capability only

fedepaol:
Some devices only output the capability (like hardware-receive)
without specifying the type (HW or SW). In this case we treat the row as
not existing.

Sample output:
ethtool -T ens3f1
Time stamping parameters for ens3f1:
Capabilities:
	hardware-transmit
	software-transmit
	hardware-receive
	software-receive
	software-system-clock
	hardware-raw-clock

Signed-off-by: Federico Paolinelli <fpaoline@redhat.com>

### DIFF
--- a/test/ptp/ptp.go
+++ b/test/ptp/ptp.go
@@ -530,13 +530,13 @@ func isPTPEnabled(ethToolOutput *bytes.Buffer) bool {
 		line := strings.TrimPrefix(scanner.Text(), "\t")
 		parts := strings.Fields(line)
 		if parts[0] == ETHTOOL_HARDWARE_RECEIVE_CAP {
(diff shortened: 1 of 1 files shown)


test/junit.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SchSeba, line 1: please remove this one

SchSeba (commented):
just one nit

test/junit.xml
fedepaol: A great classic.
Done

SchSeba (approved):
/lgtm
/approve

PR #112, opened 2021-05-31
Bug 1957748: Add cpu and memory requests for the operator

oribon:


SchSeba:
@zshi-redhat PTAL

looks good from my side

PR #113, opened 2021-06-07
Bug 1957772: Fix openshift conformance test

oribon:
https://bugzilla.redhat.com/show_bug.cgi?id=1957772
https://bugzilla.redhat.com/show_bug.cgi?id=1960577

### DIFF
--- a/bindata/linuxptp/ptp-daemon.yaml
+++ b/bindata/linuxptp/ptp-daemon.yaml
@@ -12,6 +12,8 @@ spec:
       app: linuxptp-daemon
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 33%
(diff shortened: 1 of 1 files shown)


bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -12,6 +12,8 @@ spec:
       app: linuxptp-daemon
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 33%
```
SchSeba, line 16: why you use 33% and not 1 for example any special reason?

knobunc: 1 makes upgrades very slow... 10% is what most things do, but components choose something larger.

PR #116, opened 2021-06-16
Fix setting CGO_ENABLED in build

josephdrichard:
Using CGO_ENABLED to allow building on fedora 34.  Not currently able to pass it in, and need to modify this build script if I want to override.


SchSeba (commented):
/ok-to-test
/lgtm
/approve

PR #119, opened 2021-07-06
PTP Operator with Event Sidecar

aneeshkp:
Note: This PR has a dependency on  PTP Linux daemon PR:  https://github.com/openshift/linuxptp-daemon/pull/39
This  PR addresses PTP operator plug-ins for fast event notification sidecar enablement.
This is one of the two PR to enable  PTP operator with the Event framework; to be deployed as a sidecar container, where sidecar container parses logs from LINUX PTP daemon via Unix socket and generates metrics and PTP events and transport to subscribers via AMQP bus . The event publisher is configured via  PtpOperatorConfig CRD.
Bu default PTP Operator will deploy PTP Linux daemonset without event side car enabled.
To enable the event framework, you need to set the `enableEventPublisher` field and other fields under the ptpEventConfig spec.

NOTE: This PR does not address deployment manifests and image references to new images.  The sidecar event is still in the process of getting productized and will be available soon. The next  PR will update CSV and images reference to event sidecar downstream repo.
```
apiVersion: ptp.openshift.io/v1
kind: PtpOperatorConfig
metadata:
  name: default
spec:
  daemonNodeSelector:
    feature.node.kubernetes.io/ptp-capable: "yes"
    node-role.kubernetes.io/worker: ""
  ptpEventConfig:
    enableEventPublisher: true
    holdOverTimeout: 10
    maxOffsetThreshold: 100
    minOffsetThreshold: -100
    transportHost: "amqp://amq-interconnect.amqp-interconnect.svc.cluster.local"

```


hack/env.sh
```diff
@@ -9,6 +9,9 @@ export OPERATOR_NAME=ptp-operator
 LINUXPTP_DAEMON_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-ptp | jq --raw-output '.Digest')
 PTP_OPERATOR_IMAGE_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-ptp-operator | jq --raw-output '.Digest')
 KUBE_RBAC_PROXY_DIGEST=$(skopeo inspect docker://quay.io/openshift/origin-kube-rbac-proxy | jq --raw-output '.Digest')
+SIDECAR_EVENT_DIGEST="latest"
+
 export LINUXPTP_DAEMON_IMAGE=${LINUXPTP_DAEMON_IMAGE:-quay.io/openshift/origin-ptp@${LINUXPTP_DAEMON_IMAGE_DIGEST}}
 export PTP_OPERATOR_IMAGE=${PTP_OPERATOR_IMAGE:-quay.io/openshift/origin-ptp-operator@${PTP_OPERATOR_IMAGE_DIGEST}}
 export KUBE_RBAC_PROXY_IMAGE=${KUBE_RBAC_PROXY_IMAGE:-quay.io/openshift/origin-kube-rbac-proxy@${KUBE_RBAC_PROXY_DIGEST}}
+export SIDECAR_EVENT_IMAGE=${SIDECAR_EVENT_IMAGE:-quay.io/aneeshkp/cloud-event-proxy@${SIDECAR_EVENT_DIGEST}}
```
josephdrichard, line 17: Should this be updated before merging?

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
```diff
@@ -14,6 +14,7 @@ type PtpOperatorConfigSpec struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
 	DaemonNodeSelector map[string]string `json:"daemonNodeSelector"`
+	EventConfig PtpEventConfig `json:"ptpEventConfig,omitempty"`
```
SchSeba, line 17: I think you need to remove the omitempty here

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
```diff
@@ -201,6 +202,26 @@ func (r *ReconcilePtpOperatorConfig) syncLinuxptpDaemon(defaultCfg *ptpv1.PtpOpe
 	data.Data["Namespace"] = names.Namespace
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["KubeRbacProxy"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
+	data.Data["SideCar"] = os.Getenv("SIDECAR_EVENT_IMAGE")
+	// configure EventConfig
+	data.Data["EnableEventPublisher"] = "false"
+	data.Data["EventTransportHost"] = defaultCfg.Spec.EventConfig.TransportHost
```
SchSeba, line 208: will this work if the default (not evenConfig exist)?

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
```diff
@@ -201,6 +202,26 @@ func (r *ReconcilePtpOperatorConfig) syncLinuxptpDaemon(defaultCfg *ptpv1.PtpOpe
...
+	data.Data["HoldOverTimeout"] = defaultCfg.Spec.EventConfig.HoldOverTimeout
+	data.Data["MaxOffsetThreshold"] = defaultCfg.Spec.EventConfig.MaxOffsetThreshold
+	data.Data["MinOffsetThreshold"] = defaultCfg.Spec.EventConfig.MinOffsetThreshold
+	if defaultCfg.Spec.EventConfig.EnableEventPublisher {
+		data.Data["EnableEventPublisher"] = "true"
+		if defaultCfg.Spec.EventConfig.HoldOverTimeout == 0 {
+			data.Data["HoldOverTimeout"] = 5 //min 5 secs
```
SchSeba, line 215: lets have a info log pointing out what is the time we use and why(no config provided)

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
```diff
@@ -201,6 +202,26 @@ func (r *ReconcilePtpOperatorConfig) syncLinuxptpDaemon(defaultCfg *ptpv1.PtpOpe
...
+		}
+		if defaultCfg.Spec.EventConfig.MaxOffsetThreshold == 0 {
+			data.Data["MaxOffsetThreshold"] = 100 //min 100 nano secs
```
SchSeba, line 218: same

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
```diff
@@ -24,6 +25,21 @@ type PtpOperatorConfigStatus struct {
...
+type PtpEventConfig struct {
+	//EnableEventPublisher will deploy event proxy as a sidecar
+	EnableEventPublisher  bool `json:"enableEventPublisher,omitempty"`
+	//clock state to stay in holdover state in secs
+	HoldOverTimeout int64 `json:"holdOverTimeout,omitempty"`
+	//max offset in nano secs
+	MaxOffsetThreshold int64 `json:"maxOffsetThreshold,omitempty"`
+	//min offset in nano secs
+	MinOffsetThreshold int64 `json:"minOffsetThreshold,omitempty"`
+    // transportHost address for event messages
```
SchSeba, line 39: same

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
```diff
@@ -24,6 +25,21 @@ type PtpOperatorConfigStatus struct {
...
+// PtpEventConfig defines the desired state of event framework
+// +k8s:openapi-gen=true
+type PtpEventConfig struct {
+	//EnableEventPublisher will deploy event proxy as a sidecar
+	EnableEventPublisher  bool `json:"enableEventPublisher,omitempty"`
+	//clock state to stay in holdover state in secs
+	HoldOverTimeout int64 `json:"holdOverTimeout,omitempty"`
+	//max offset in nano secs
+	MaxOffsetThreshold int64 `json:"maxOffsetThreshold,omitempty"`
+	//min offset in nano secs
```
SchSeba, line 37: same

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
```diff
@@ -24,6 +25,21 @@ type PtpOperatorConfigStatus struct {
...
 }
 
+// PtpEventConfig defines the desired state of event framework
+// +k8s:openapi-gen=true
+type PtpEventConfig struct {
+	//EnableEventPublisher will deploy event proxy as a sidecar
+	EnableEventPublisher  bool `json:"enableEventPublisher,omitempty"`
+	//clock state to stay in holdover state in secs
+	HoldOverTimeout int64 `json:"holdOverTimeout,omitempty"`
+	//max offset in nano secs
```
SchSeba, line 35: same

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
```diff
@@ -24,6 +25,21 @@ type PtpOperatorConfigStatus struct {
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
 }
 
+// PtpEventConfig defines the desired state of event framework
+// +k8s:openapi-gen=true
+type PtpEventConfig struct {
+	//EnableEventPublisher will deploy event proxy as a sidecar
+	EnableEventPublisher  bool `json:"enableEventPublisher,omitempty"`
+	//clock state to stay in holdover state in secs
```
SchSeba, line 33: lets add to the description the default time we use if the user doesn't provide one

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
aneeshkp: Yes, the struct  will have empty data,  The default CRD does not have EventConfig and deploys without using EventConfig

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
aneeshkp: I can add those data to CR status .So you always know the config status. Will that work?

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
aneeshkp:
```
apiVersion: ptp.openshift.io/v1
kind: PtpOperatorConfig
metadata:
  name: default
spec:
  daemonNodeSelector:
    feature.node.kubernetes.io/ptp-capable: "yes" 
  #ptpEventConfig: {}
```

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
aneeshkp: I made this data available via CLI 

$oc get ptpoperatorconfig
```
NAME      EVENT ENABLED   HOLDOVER TIMEOUT   MAX THRESHOLD   MIN THRESHOLD   EVENT TRANSPORT HOST
default   false           5                  100             -100            
```

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
aneeshkp: I made this data available via CLI 

$oc get ptpoperatorconfig
```
NAME      EVENT ENABLED   HOLDOVER TIMEOUT   MAX THRESHOLD   MIN THRESHOLD   EVENT TRANSPORT HOST
default   false           5                  100             -100            
```

pkg/controller/ptpoperatorconfig/ptpoperatorconfig_controller.go
aneeshkp: I made this data available via CLI 

$oc get ptpoperatorconfig
```
NAME      EVENT ENABLED   HOLDOVER TIMEOUT   MAX THRESHOLD   MIN THRESHOLD   EVENT TRANSPORT HOST
default   false           5                  100             -100            
```

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
aneeshkp: Also added default  value so the CRD's show the default value
I made this data available via CLI 

$oc get ptpoperatorconfig
```
NAME      EVENT ENABLED   HOLDOVER TIMEOUT   MAX THRESHOLD   MIN THRESHOLD   EVENT TRANSPORT HOST
default   false           5                  100             -100            
```

aneeshkp:
rebased

api/v1/ptpoperatorconfig_types.go
```diff
@@ -58,6 +65,24 @@ type PtpOperatorConfigList struct {
...
+	// clock state to stay in holdover state in secs
+	HoldOverTimeout int64 `json:"holdOverTimeout,omitempty"`
+	// +kubebuilder:default=100
+	// max offset in nano secs
+	MaxOffsetThreshold int64 `json:"maxOffsetThreshold,omitempty"`
+	// +kubebuilder:default=-100
+	// min offset in nano secs
+	MinOffsetThreshold int64 `json:"minOffsetThreshold,omitempty"`
+	// transportHost address for event messages.Example amqp://<service-name>.<namespace>.svc.cluster.local
+	TransportHost string `json:"transportHost,omitempty"`
```
SchSeba, line 83: will this work if the transport is empty? if not we need to remove the omitempty

bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -23,63 +23,106 @@ spec:
...
+            - "--api-port=8080"
+          volumeMounts:
+            - name: pubsubstore
+              mountPath: /store
+            - name: logs-socket-dir
+              mountPath: /tmp
+          ports:
+            - name: metrics-port
+              containerPort: 9091
+          env:
```
SchSeba, line 50: I think better to have all this in a config map that the event sidecar will use. this way in the future if we need other fields there is no need to change the ptp operator just the sidecar code

api/v1/ptpoperatorconfig_types.go
```diff
@@ -58,6 +65,24 @@ type PtpOperatorConfigList struct {
 	Items           []PtpOperatorConfig `json:"items"`
 }
 
+// PtpEventConfig defines the desired state of event framework
+type PtpEventConfig struct {
```
SchSeba, line 69: is it possible to have in one cluster different thresholds for different nodes?

if that is the case I think we can have here just a flag to enable disable the event bus sidecar and move the thresholds as part of the ptp-config so it will be applied with the policy.

In general I think the thresholds depend on the policy you use so they need to be there

bundle/manifests/ptp.openshift.io_ptpoperatorconfigs.yaml
```diff
@@ -14,7 +14,28 @@ spec:
     singular: ptpoperatorconfig
   scope: Namespaced
   versions:
-  - name: v1
+  - additionalPrinterColumns:
```
SchSeba, line 17: there is a reason to print all of this when the use do `oc get ptpoperatorconfig` ?

maybe just the even enabled status

config/manager/env.yaml
```diff
@@ -16,4 +16,6 @@ spec:
             - name: LINUXPTP_DAEMON_IMAGE
               value: "quay.io/openshift/origin-ptp:4.9"
             - name: KUBE_RBAC_PROXY_IMAGE
-              value: "quay.io/openshift/origin-kube-rbac-proxy:4.9"
\ No newline at end of file
+              value: "quay.io/openshift/origin-kube-rbac-proxy:4.9"
+            - name: SIDECAR_EVENT_IMAGE
+              value: "quay.io/openshift/cloud-event-proxy:4.9"
```
SchSeba, line 21: please add a new line here

config/crd/bases/ptp.openshift.io_ptpoperatorconfigs.yaml
```diff
@@ -16,7 +16,28 @@ spec:
     singular: ptpoperatorconfig
```
SchSeba, line 16: you also need to change this in the manifests folder

api/v1/ptpoperatorconfig_types.go
aneeshkp: it can be empty if EnableEventPublisher=false. Also if it is empty, the operator will throw an error in the log and it will not deploy the pod.

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: This is the same as LinuxPTPDaemon. I am just moving metrics from LinuxPTPDaemon to Sidecar when EventPublisher is enabled. trying to keep it consistent with the metric endpoint when we move it to the sidecar

pkg/apis/ptp/v1/ptpoperatorconfig_types.go
aneeshkp: Added default to decriptioin

api/v1/ptpoperatorconfig_types.go
aneeshkp: @SchSeba  it's possible, not likely. But there is a lot of overhead of providing this functionality at this point., i.e. moving this to ptp-config.
if I move this to ptp-config then sidecar needs to watch for ptp-config changes in order to know these values before it can process logs coming from Linux PTP-daemon. That means either the Linux ptp daemon has to send me these values or I need to read the .conf files periodically.

aneeshkp: I think 4.9 is primarily SNO, I can do this in the next release and capture this feedback in the user story?

bundle/manifests/ptp.openshift.io_ptpoperatorconfigs.yaml
aneeshkp: ok, I will just print the event enabled here.

api/v1/ptpoperatorconfig_types.go
SchSeba: The problem is the API is v1 this means we will need to support this for a long time before we are able to depricate.

we already pass all the information to the linuxptp using a configmap you can do the same as the linuxptp is doing right now

api/v1/ptpoperatorconfig_types.go
SchSeba: so maybe we don't even need the EnableEventPublisher

lets just make :
EventConfig *PtpEventConfig `json:"ptpEventConfig,omitempty"`

and if it's not nil all the variables most exist

bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -23,63 +23,106 @@ spec:
...
+            - name: linuxptp-certs
+              mountPath: /etc/metrics
+              readOnly: True
+        - name: linuxptp-daemon-container
+          securityContext:
+            privileged: true
+          image: {{.Image}}
+          command: [ "/bin/bash", "-c", "--" ]
+          args: [ "/usr/local/bin/ptp --alsologtostderr -v 10" ]
+          imagePullPolicy: Always
```
SchSeba, line 92: why there was a need to change this?

bindata/linuxptp/ptp-daemon.yaml
SchSeba: what you mean here can you please explain me again?

I think you can just use 

```
 - name: config-volume
          configMap:
            name: ptp-configmap
```

and use it to get the informantion you need for the sidecard

bundle/manifests/ptp.openshift.io_ptpoperatorconfigs.yaml
SchSeba: please take a look on the api change request I don't think we need the event enabled at all

api/v1/ptpoperatorconfig_types.go
aneeshkp: The issue is sidecar is already running, ptp config can(will)   be applied after the sidecar is in running. So that mean sidecar should be able to read configmap changes ? is that correct

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: So by default, you have the same metrics configured for daemon right, and then we have kube-rbac-proxy configure to point to 127.0.0.1:9091. Putting in configmap and allowing the user to change metrics endpoint, will also need to change kube-rbac-proxy? is that correct? I don't see any value in changing this to config map.

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: oops! That was during testing , I will revert it back to IfNotPresent

api/v1/ptpoperatorconfig_types.go
```diff
@@ -29,6 +29,8 @@ type PtpOperatorConfigSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
 	DaemonNodeSelector map[string]string `json:"daemonNodeSelector"`
+	// +kubebuilder:default={}
+	EventConfig PtpEventConfig `json:"ptpEventConfig,omitempty"`
```
SchSeba, line 33: please make this a pointer

bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -23,63 +23,102 @@ spec:
...
-        - name: socket-dir
-          mountPath: /var/run
+       {{ if (eq .EnableEventPublisher true) }}
+        - name: cloud-native-event-sidecar
+          image: {{ .SideCar }}
+          args:
+            - "--metrics-addr=127.0.0.1:9091"
+            - "--store-path=/store"
+            - "--transport-host={{ .EventTransportHost }}"
+            - "--api-port=8080"
```
SchSeba, line 41: general question: what is this port?

bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -23,63 +23,102 @@ spec:
...
+            - name: LOGS_TO_SOCKET
+              value: "{{ .EnableEventPublisher }}"
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/linuxptp
+            - name: socket-dir
+              mountPath: /var/run
+            {{ if (eq .EnableEventPublisher true) }}
+            - name: logs-socket-dir
+              mountPath: /tmp
```
SchSeba, line 107: nit: maybe a better folder like even-bus-sockets

bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -23,63 +23,102 @@ spec:
...
+            {{ end }}
       volumes:
         - name: config-volume
           configMap:
             name: ptp-configmap
         - name: linuxptp-certs
           secret:
             secretName: linuxptp-daemon-secret
+        {{ if (eq .EnableEventPublisher true) }}
+        - name: pubsubstore
```
SchSeba, line 117: question: is the emptyDir fine here? it will be removed if the pod is deleted that is expected?
maybe you need a hostPath here?

api/v1/ptpoperatorconfig_types.go
```diff
@@ -58,6 +61,15 @@ type PtpOperatorConfigList struct {
 	Items           []PtpOperatorConfig `json:"items"`
 }
 
+// PtpEventConfig defines the desired state of event framework
+type PtpEventConfig struct {
+	// +kubebuilder:default=false
+	// EnableEventPublisher will deploy event proxy as a sidecar
+	EnableEventPublisher bool `json:"enableEventPublisher,omitempty"`
+	// transportHost address for event messages.Example amqp://<service-name>.<namespace>.svc.cluster.local
+	TransportHost string `json:"transportHost,omitempty"`
```
SchSeba, line 70: should you have a port here also?

api/v1/ptpconfig_types.go
```diff
@@ -66,6 +66,20 @@ type PtpProfile struct {
 	Ptp4lOpts   *string `json:"ptp4lOpts,omitempty"`
 	Phc2sysOpts *string `json:"phc2sysOpts,omitempty"`
 	Ptp4lConf   *string `json:"ptp4lConf,omitempty"`
+	// +kubebuilder:default={}
```
SchSeba, line 69: I think you can remove this it can be just nil

config/manager/env.yaml
```diff
@@ -16,4 +16,7 @@ spec:
...
```
SchSeba, line 21: I am not able to pull this image

podman pull quay.io/openshift/cloud-event-proxy:4.9
Error response from daemon: unauthorized: access to the requested resource is not authorized

api/v1/ptpoperatorconfig_types.go
```diff
@@ -29,6 +29,8 @@ type PtpOperatorConfigSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
 	DaemonNodeSelector map[string]string `json:"daemonNodeSelector"`
+	// +kubebuilder:default={}
```
SchSeba, line 32: I think you can remove this it can be just nil

SchSeba (commented):
looks good now just some small comments and we can merge it.

the important point is the sidecard image doesn't exist we can't merge without it

api/v1/ptpoperatorconfig_types.go
aneeshkp: The sidecar rest api port is not that dynamic. The port is an internal sidecar rest  API that is only accessible by the ptp-operator-plugin via localhost only. it is not used by anything else.

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: Sidecar runs an API service based on ORAN API specifications, The service is used to create publisher/subscription and post and receive events to and from the framework

config/manager/env.yaml
aneeshkp: it's not yet ready. Nishant is working on building this image, right now it's in a private repo. This PR cannot be merged till that repo is ready

api/v1/ptpconfig_types.go
aneeshkp: having default allows me to populate with default values. Or users will not know the default values unless they look for logs.

aneeshkp: time="2021-07-19T19:13:33Z" level=info msg="update ptp threshold values for ens5f1\n holdoverTimeout: 5\n maxThreshold: 100\n minThreshold: -100\n"

aneeshkp: removed the default and checking for nil in sidecar

api/v1/ptpoperatorconfig_types.go
aneeshkp: Fixed

aneeshkp:
need review openshift/linuxptp-daemon#39 related to this PR

api/v1/ptpoperatorconfig_types.go
aneeshkp: fixed

api/v1/ptpoperatorconfig_types.go
aneeshkp: fixed

api/v1/ptpoperatorconfig_types.go
aneeshkp: fixed: converted to a pointer

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: yes, it is expected behavior, if the pod is deleted then the store is emptied too, but available if one of the container restarts.

api/v1/ptpoperatorconfig_types.go
aneeshkp: Fixed: Moved to ptpConfig

api/v1/ptpoperatorconfig_types.go
aneeshkp: amqp transportHost doesn't have a port, we just use service name like `amqp://<service-name>.<namespace>.svc.cluster.local`

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: Fixed: renamed to event-bus-socket

aneeshkp:
@SchSeba https://github.com/openshift/linuxptp-daemon/pull/39 depends on this PR

api/v1/ptpoperatorconfig_types.go
SchSeba: I can still see it :P

bindata/linuxptp/ptp-daemon.yaml
SchSeba: My problem was with the folder here "/tmp" I think we can have a better folder

controllers/ptpoperatorconfig_controller.go
```diff
@@ -172,6 +172,19 @@ func (r *PtpOperatorConfigReconciler) syncLinuxptpDaemon(defaultCfg *ptpv1.PtpOp
 	data.Data["Namespace"] = names.Namespace
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASEVERSION")
 	data.Data["KubeRbacProxy"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
+	data.Data["SideCar"] = os.Getenv("SIDECAR_EVENT_IMAGE")
+	// configure EventConfig
+	if defaultCfg.Spec.EventConfig==nil{
```
SchSeba, line 177: you should run `go fmt`

SchSeba:
Hi @aneeshkp please update this file also

https://github.com/openshift/ptp-operator/blob/master/manifests/4.9/image-references

SchSeba:
@aneeshkp please rebase the PR

SchSeba:
This looks good just squash all the commits please

api/v1/ptpoperatorconfig_types.go
aneeshkp: fixed

controllers/ptpoperatorconfig_controller.go
aneeshkp: fixed

PR #127, opened 2021-07-29
Bug 1987136: Declare supported arches in CSV

yselkowitz:
The OperatorHub currently assumes x86-only for any operators which do
not declare any such labels.  Declaring these explicitly should make the
current status more obvious in the code, and make it easier to correctly
add further architectures when conditions warrant.


yselkowitz:
Is anyone watching PRs in this repo??
/assign @fedepaol

yselkowitz:
I did not plan to backport, this is the assumed default if unspecified, but making it explicit is clearer and easier to understand how to change in the future if/when more arches are supported.

PR #131, opened 2021-08-15
Bug 1993854: Added /var/run volume for sidecar

aneeshkp:
This PR adds /var/run volume  for cloud-event-proxy sidecar  for event management.
The new multi-interface changes to  Linux-ptp-daemon and  cloud-event-proxy , requires cloud-event-proxy to watch for ptp4l.X.config file updates read interface details.


SchSeba:
@aneeshkp please add the bug here so it will have the BZ valid label

PR #132, opened 2021-08-26
Bug 1998598: added ptp event config manifest to 4.9

aneeshkp:
4.9 manifest files had missing spec fields for ptpEventConfig.
This is required for the operator to work with new ptp event config


SchSeba:
@aneeshkp you also need to update the ptpCofig CRD in the manifest folder

aneeshkp:
@SchSeba , Thanks  I added  ptpConfig

PR #133, opened 2021-08-31
Bug 1999886:fixed bind address issue

aneeshkp:
Sidecar API was configured to listen to 8080, although sidecar was designed  to run with hostnetwork=false, but ptp-operator requires hostNetwork set to true , hence 8080 was giving bind address already in use.
Changed port to 8089 to avoid any address in use error.


SchSeba (commented):
/lgtm
/approve

PR #135, opened 2021-09-02
add host ip env

aneeshkp:
Event side car needs to know host _ip to construct amqp transport that was created as nodeport service
access as amq://nodeip:port
This PR passes node_ip to the event sidecar


SchSeba:
@aneeshkp do you need this to be backported to 4.9, if that is the case please open a BZ

PR #148, opened 2021-11-04
Fixup ptp-operator bundle

josephdrichard:
Remove serviceaccount manifests from bundle, as they conflicted with
deploy.  Also rebuild bundle.


josephdrichard:
Addressed by https://github.com/openshift/ptp-operator/pull/153

PR #149, opened 2021-11-06
Add ptp scheduling to PtpProfile crd

josephdrichard:
Add PtpSchedulingPolicy and PtpSchedulingPriority to PtpProfile crd.
These will be used to allow running ptp4l and phc2sys with custom
priority (e.g. SCHED_FIFO).

### DIFF
--- a/bundle/manifests/ptp.openshift.io_ptpconfigs.yaml
+++ b/bundle/manifests/ptp.openshift.io_ptpconfigs.yaml
@@ -65,6 +65,11 @@ spec:
                           format: int64
                           type: integer
                       type: object
+                    ptpSchedulingPolicy:
+                      type: string
+                    ptpSchedulingPriority:
+                      format: int64
+                      type: integer
                   required:
                   - name
                   type: object
--- a/config/crd/bases/ptp.openshift.io_ptpconfigs.yaml
+++ b/config/crd/bases/ptp.openshift.io_ptpconfigs.yaml
@@ -67,6 +67,11 @@ spec:
                           format: int64
                           type: integer
                       type: object
+                    ptpSchedulingPolicy:
+                      type: string
+                    ptpSchedulingPriority:
+                      format: int64
+                      type: integer
                   required:
                   - name
                   type: object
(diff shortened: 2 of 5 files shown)


josephdrichard:
@aneeshkp @SchSeba

williamcaban:
The PTP Operator should work with RT and non-RT kernels. If this is running as part of the housekeeping CPUs, using low latency scheduling may have unintended consequences by starving other system processes which might be already constraints.

Consider adjusting the scheduling priority for Pod PID and the scheduling priority of the relevant processes inside the Pod over providing a way to modify the scheduling policy which opens the node to higher constraints of resources by user misconfiguration of those parameters.

josephdrichard:
> The PTP Operator should work with RT and non-RT kernels. If this is running as part of the housekeeping CPUs, using low latency scheduling may have unintended consequences by starving other system processes which might be already constraints.
> 
> Consider adjusting the scheduling priority for Pod PID and the scheduling priority of the relevant processes inside the Pod over providing a way to modify the scheduling policy which opens the node to higher constraints of resources by user misconfiguration of those parameters.

There are use cases where it's necessary for a user to be able to configure this.  In order to prevent drowning out other processes, there will be some validation done on input, however if a user does not want this to be changes, they don't need to.

bundle/manifests/ptp.openshift.io_ptpconfigs.yaml
```diff
@@ -65,6 +65,10 @@ spec:
                           format: int64
```
SchSeba, line 65: please copy this file to the manifests/stable folder

config/crd/bases/ptp.openshift.io_ptpconfigs.yaml
```diff
@@ -67,6 +67,10 @@ spec:
                           format: int64
```
SchSeba, line 67: please copy this file to the manifests/stable folder

SchSeba:
we talked offline I will merge this PR to allow the vendoring in the linuxptp-daemon.

I will create an assign an issue to @josephdrichard to finish the needed work here

/lgtm
/approve

PR #153, opened 2021-11-14
Fix webhook issue

SchSeba:
This commit fix the webhook issue introduced in https://github.com/openshift/ptp-operator/pull/140

This also fix the make bundle validation error.

And last bump the k8s version to 1.22.2 with latest operator-sdk 1.14.0


config/webhook/manifests.yaml
```diff
@@ -14,8 +12,8 @@ webhooks:
       name: ptpconfig-validator-webhook-service
       namespace: system
       path: /validate-ptp-openshift-io-v1-ptpconfig
-  failurePolicy: Ignore
-  name: vptpconfig.kb.io
+  failurePolicy: Fail
```
josephdrichard, line 15: If the webhook is not running in the ptp-operator, it should default to accept in my opinion
failurePolicy is specifically for the case that the webhook fails(e.g. not in ptp operator pod yet), not if the webhook rejects an input.
Specifically for the case that the ptp-operator pod isn't up and running, I would lean to allowing creating a ptp-config still.
As this changes behaviour, if this change is required, it should also be in a different PR in my opniion a

josephdrichard (changes requested):

josephdrichard (approved):
/lgtm

PR #155, opened 2021-11-17
Remove the Z stream

SchSeba:
This is needed as we don't use Z stream here


SchSeba:
ci issues not related to the PR reruning 

/retest

PR #159, opened 2021-11-23
Add validation of scheduling policy

josephdrichard:


josephdrichard:
@SchSeba @aneeshkp

api/v1/ptpconfig_types.go
```diff
@@ -61,12 +61,15 @@ type PtpConfigList struct {
...
-	Ptp4lOpts             *string            `json:"ptp4lOpts,omitempty"`
-	Phc2sysOpts           *string            `json:"phc2sysOpts,omitempty"`
-	Ptp4lConf             *string            `json:"ptp4lConf,omitempty"`
-	PtpSchedulingPolicy   *string            `json:"ptpSchedulingPolicy,omitempty"`
+	Name        *string `json:"name"`
+	Interface   *string `json:"interface,omitempty"`
+	Ptp4lOpts   *string `json:"ptp4lOpts,omitempty"`
+	Phc2sysOpts *string `json:"phc2sysOpts,omitempty"`
+	Ptp4lConf   *string `json:"ptp4lConf,omitempty"`
+	// +kubebuilder:validation:Enum=SCHED_OTHER;SCHED_FIFO;
```
SchSeba, line 69: I think `SCHED_OTHER` is not valid here as there is nothing in the linuxptp

api/v1/ptpconfig_webhook.go
```diff
@@ -118,6 +122,12 @@ func (r *PtpConfig) validate() error {
 		if summary_interval != logSyncInterval {
 			return errors.New("summary_interval " + summary_interval + " must match logSyncInterval " + logSyncInterval)
 		}
+
+		if profile.PtpSchedulingPolicy != nil && *profile.PtpSchedulingPolicy == "SCHED_FIFO" {
```
SchSeba, line 126: do you need to check also about `SCHED_OTHER`?

SchSeba (commented):
can you please rebase on master we already merge one of the commits from this PR on a different one

api/v1/ptpconfig_types.go
josephdrichard: SCHED_OTHER will have the same behaviour as just not setting it.

api/v1/ptpconfig_webhook.go
josephdrichard: My opinion is that we don't.  If SCHED_OTHER, no validation is necessary.

PR #164, opened 2021-12-05
Add test case to verify SCHED_FIFO propagated to linuxptp-daemon

josephdrichard:
Skip if no ptpconfig with profile.PtpSchedulingPolicy=="SCHED_FIFO"
Fail if "/bin/chrt -f %d /usr/sbin/ptp4l" missing from daemon logs.

### DIFF
--- a/test/conformance/ptp/ptp.go
+++ b/test/conformance/ptp/ptp.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -121,6 +123,7 @@ var _ = Describe("[ptp]", func() {
 		var masterNodeLabel, slaveNodeLabel string
 		discoveryFailed := false
 		var masterProfile, slaveProfile string
+		var fifoPriorities map[string]int64
 
 		execute.BeforeAll(func() {
 			if isSingleNode {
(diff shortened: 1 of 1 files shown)


josephdrichard:
@aneeshkp

test/conformance/ptp/ptp.go
```diff
@@ -364,6 +365,54 @@ var _ = Describe("[ptp]", func() {
 				Expect(slavePodDetected).ToNot(BeFalse(), "No slave pods detected")
 			})
 		})
+
+		Context("Running with fifo scheduling", func() {
+			BeforeEach(func() {
+				if discoveryFailed {
+					Skip("Failed to find a valid ptp slave configuration")
+				}
+				ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
```
aneeshkp, line 374: You should run this only on slave pods, right?
if podRole(pod, slaveNodeLabel) {

josephdrichard: The test isn't filtering whether the ptpconfig is a master or slave config, so should run on all ptp nodes.

aneeshkp: I am not sure, I will try to run this and see but I think the below code doesn't filter non-ptp enabled pods.
ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})

test/conformance/ptp/ptp.go
```diff
@@ -364,6 +365,50 @@ var _ = Describe("[ptp]", func() {
...
+		Context("Running with fifo scheduling", func() {
+			BeforeEach(func() {
+				if discoveryFailed {
+					Skip("Failed to find a valid ptp slave configuration")
+				}
+
+				masterConfigs, slaveConfigs := discoveryPTPConfiguration(PtpLinuxDaemonNamespace)
+				ptpConfigs := append(masterConfigs, slaveConfigs...)
+
+				fifoPriorities = make(map[string]int64)
```
aneeshkp, line 378: You should add SCHED_FIFO by default to the ptpconfig , if not it can only work in discovery mode.
another option is to add an env variable to enable this as I did for PTP events @SchSeba  what are your thoughts on this?

test/conformance/ptp/ptp.go
aneeshkp: I was able to test this, marking as resoved

test/conformance/ptp/ptp.go
josephdrichard: If I were to add SCHED_FIFO to the default config, and move the test to that section mode section, then this test case would only work in a non-SNO configuration, which is not ideal either.
At some point, we do need to clean this up so that all of the test cases support both discovery and non-discovery, and both SNO and multi-node, at which point this should be moved to that section, and fifo scheduling added.  In my opinion right now though, it is more important to be able to run in SNO than to be able to run in non-discovery mode.

test/conformance/ptp/ptp.go
```diff
@@ -364,6 +365,54 @@ var _ = Describe("[ptp]", func() {
...
+			BeforeEach(func() {
+				if discoveryFailed {
+					Skip("Failed to find a valid ptp slave configuration")
+				}
+				ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(ptpPods.Items)).To(BeNumerically(">", 0), fmt.Sprint("linuxptp-daemon is not deployed on cluster"))
+
+				for _, pod := range ptpPods.Items {
+					ptpRunningPods = append(ptpRunningPods, pod)
```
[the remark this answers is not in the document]
josephdrichard: updated to just get pods when needed.

josephdrichard:
Tests looked good.
/hold cancel

PR #165, opened 2021-12-06
Add ptp unit test for events

aneeshkp:
Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>
Adds following unit test for testing PTP event
1. Enable ptp event by setting env variable ENABLE_PTP_EVENT
2. Verifies by checking the following 
  - Checking event side car is present
  - Checking event metrics are present
  - Checking  event api is healthy
  - Checking events are generated

### DIFF
--- a/test/conformance/ptp/ptp.go
+++ b/test/conformance/ptp/ptp.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/ptp-operator/test/utils/event"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
@@ -45,6 +47,16 @@ var _ = Describe("[ptp]", func() {
 
 	Context("PTP configuration verifications", func() {
 		// Setup verification
+		// if requested enabled  ptp events
+		It("Should check whether PTP operator needs to enable PTP events", func() {
+			By("Find if variable set to enable ptp events")
+			if event.Enable() {
+				Expect(enablePTPEvent()).NotTo(HaveOccurred())
+				ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), "default", metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(ptpConfig.Spec.EventConfig.EnableEventPublisher).Should(BeTrue(), "failed to enable ptp event")
+			}
+		})
 		It("Should check whether PTP operator appropriate resource exists", func() {
 			By("Getting list of available resources")
 			rl, err := client.Client.ServerPreferredResources()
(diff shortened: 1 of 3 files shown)


aneeshkp:
@josephdrichard @SchSeba

test/conformance/ptp/ptp.go
```diff
@@ -356,14 +370,102 @@ var _ = Describe("[ptp]", func() {
...
+					Skip("Failed to find a valid ptp slave configuration")
+				}
+				ptpPods, err := client.Client.Pods(PtpLinuxDaemonNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=linuxptp-daemon"})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(ptpPods.Items)).To(BeNumerically(">", 0), "linuxptp-daemon is not deployed on cluster")
+
+				for _, pod := range ptpPods.Items {
+					if podRole(pod, slaveNodeLabel) {
+						waitUntilLogIsDetected(pod, 3*time.Minute, "Profile Name:")
+						ptpSlaveRunningPods = append(ptpSlaveRunningPods, pod)
```
josephdrichard, line 401: any reason for this to not run on ptp masters?

test/conformance/ptp/ptp.go
```diff
@@ -356,14 +370,102 @@ var _ = Describe("[ptp]", func() {
...
+					Expect(len(ptpSlaveRunningPods)).To(BeNumerically(">=", 1), "Fail to detect PTP slave pods on Cluster")
+				}
+			})
+
+			It("Should check for ptp events ", func() {
+				By("Checking event side car is present")
+
+				for _, pod := range ptpSlaveRunningPods {
+					cloudProxyFound := false
+					Expect(len(pod.Spec.Containers)).To(BeNumerically("==", 3), "linuxptp-daemon is not deployed on cluster with cloud event proxy")
```
josephdrichard, line 416: Any reason this check is necessary?  This will break if adding another container at some point, right?

josephdrichard (commented):
Can you explain what you are doing in the commit message?

test/conformance/ptp/ptp.go
aneeshkp: This is valid for event proxy . Right now we have 3 containers  when event is enabled , that is we run event proxy as side car  . When event is not running we just have 2 containers .
Adding another container will be major changes which needs to revisit this and I don’t think we will be adding another sidecar to this .

test/conformance/ptp/ptp.go
aneeshkp: I ignored it because I am interested in slave configured events only and I am merely validating event functionality. 
technically one node is good test . 
Also I am not validating  the data itself , just the functionality and slave logs are more valid for my test case . 
what do you think ? Should I enhance it to include master too?

aneeshkp: Also it will be faster just checking one pod .

PR #166, opened 2021-12-13
Bug 2034484: feat: bumped library-go version

eggfoobar:
This should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242).

Changes:
- added conventional defaults for leader election
- added cluster query for leader election for SNO environments
    - at start up operator will query if it is currently in a SNO or HA topology and set LE Config respective of those environments

Signed-off-by: ehila <ehila@redhat.com>


main.go
```diff
@@ -38,6 +38,7 @@ import (
 
 	ptpv1 "github.com/openshift/ptp-operator/api/v1"
 	"github.com/openshift/ptp-operator/controllers"
+	"github.com/openshift/ptp-operator/pkg/leaderelection"
```
josephdrichard, line 41: Is the leaderelection implementation specific to ptp-operator? Why aren't you just importing https://github.com/openshift/library-go/blob/master/pkg/config/leaderelection/leaderelection.go here?

eggfoobar: Hey @josephdrichard You're right this isn't specific to ptp-operator. When the PR was opened there was a breaking dependency change between the k8s-23 libs and controller-runtime. It was causing some problems and I figured a bit of code was the better approach for now since library-go wasn't being used else where in the code. Luckily, controller-runtime was updated a few hours so we can import it now.

josephdrichard: Are you still looking to update this?

eggfoobar: Apologies, forgot to mention in the reply. This PR has been updated with the latest library-go code and I'm using those methods to drive the leaderelection configs

josephdrichard:
Was this tested on a running cluster?

eggfoobar:
> Was this tested on a running cluster?

Hey @josephdrichard, for testing this operator I have been relying on the CI cluster to validate that the correct leader election config is being applied. The correct config was applied in there, updated the rbac permissions since I had missed to generate those.

eggfoobar:
/retitle Bug 2034484: feat: bumped library-go version
This PR bumps to latest library-go to utilize changes for performance gains on SNO DU clusters

PR #183, opened 2022-05-09
adding support for external grandmaster (SNO support)

edcdavid:


test/conformance/ptp/ptp.go
```diff
@@ -126,10 +126,9 @@ var _ = Describe("[ptp]", func() {
 		var fifoPriorities map[string]int64
 
 		execute.BeforeAll(func() {
-			if isSingleNode {
-				Skip("Running in single node mode")
-			}
-			if !discovery.Enabled() {
+			// Configure PTP only if more that 1 node is available
+			// if only 1 node is present, only option is to discover config
+			if !discovery.Enabled() && !isSingleNode{
```
josephdrichard, line 131: Should we skip if discovery isn't enabled and it's a single node?

test/conformance/ptp/ptp.go
```diff
@@ -192,7 +188,7 @@ var _ = Describe("[ptp]", func() {
 						ptpMasterRunningPods = append(ptpMasterRunningPods, pod)
 					}
 				}
-				if discovery.Enabled() {
+				if discovery.Enabled() || isSingleNode {
```
josephdrichard, line 191: Why do we need to check if isSingleNode?  If it's a single node, won't discovery always be enabled?

test/conformance/ptp/ptp.go
edcdavid: I was assuming that discovery would be mandatory for single node. Will remove it.

test/conformance/ptp/ptp.go
edcdavid: same here I assumed that irrespective of the value of discovery enabled, single node means that discovery would always be enabled. will change it

test/conformance/ptp/ptp.go
aneeshkp: if discovery is not enabled, then you need to create default PTPconfig for a single node based on request either as OC or BC.

test/conformance/ptp/ptp.go
```diff
@@ -355,7 +348,7 @@ var _ = Describe("[ptp]", func() {
...
```
aneeshkp, line 351: when it is singNode , you will label node as ptp/slave only

api/v1/ptpconfig_webhook.go
```diff
@@ -136,21 +139,18 @@ func (r *PtpConfig) ValidateDelete() error {
 	return nil
 }
 
-func getInterfaces(input *ptp4lConf, mode string) (interfaces []string) {
```
josephdrichard: If you are making multiple changes to the same line, please squash your commits.  It's very difficult to review code that's making changes to a commit that hasn't been reviewed already.

api/v1/ptpconfig_webhook.go
```diff
@@ -136,21 +139,18 @@ func (r *PtpConfig) ValidateDelete() error {
...
-	modeInt := "0"
-	if mode == Master {
-		modeInt = "1"
-	}
+func getInterfaces(input *ptp4lConf, mode PtpRole) (interfaces []string) {
+
 	for index, section := range input.sections {
 		sectionName := strings.TrimSpace(strings.ReplaceAll(strings.ReplaceAll(index, "[", ""), "]", ""))
-		if strings.TrimSpace(section.options["masterOnly"]) == modeInt {
+		if strings.TrimSpace(section.options["masterOnly"]) == strconv.Itoa(int(mode)) {
```
josephdrichard, line 146: masterOnly is not a valid int, the validation webhook will crash?

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,274 @@
...
+	DiscoveryFailureStatus
+)
+
+type PTPMode int64
+
+const (
+	OrdinaryClock PTPMode = iota
+	BoundaryClock
+	DualNICBoundaryClock
+	Discovery
```
aneeshkp, line 46: Will this change exiting discovery env available?  it  is used by cn-test

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	if numBc == 2 {
+		GlobalConfig.PtpModeDiscovered = DualNICBoundaryClock
+		GlobalConfig.Status = DiscoverySuccessStatus
+	}
+
+}
+
+func IsPtpSlave(ptp4lOpts string, phc2sysOpts string) bool {
+	return strings.Contains(ptp4lOpts, "-s") && strings.Count(phc2sysOpts, "-a") == 1 && strings.Count(phc2sysOpts, "-r") == 1
```
aneeshkp, line 240: for dual ptp , in one of the slave ptpconfig (nic 2). There won't be a  phc2sysOpts metadata.

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,274 @@
...
+			slaveIf := len(ptpv1.GetInterfaces(pickedSlaveConfig.Config, ptpv1.Slave))
+			GlobalConfig.Status = DiscoveryFailureStatus
+			// OC
+			if masterIf == 0 && slaveIf == 1 {
+				GlobalConfig.PtpModeDiscovered = OrdinaryClock
+				GlobalConfig.Status = DiscoverySuccessStatus
+				GlobalConfig.DiscoveredSlavePtpConfig = pickedSlaveConfig
+			}
+			// BC and Dual NIC BC
+			if masterIf > 1 && slaveIf == 1 {
```
aneeshkp, line 215: you can have slaveIf >1 , in that case one of the slave interfaces will be in passive mode.

aneeshkp, line 215: you can have master interface==1

test/conformance/ptp/ptp.go
edcdavid: fixed

api/v1/ptpconfig_webhook.go
edcdavid: squashed the commits

api/v1/ptpconfig_webhook.go
edcdavid: "mode" is of type PtpRole int, which is an int. We should be ok. Also, I did not touch the original code at all. The reason I added the get interface function here was to reuse the code in populatePtp4lConf. I see 3 options:
- add a function in ptpconfig_webhook.go without touching the original code (current PR code)
- modify the original code to make the PopulatePtp4lConf usable in test code
- duplicate the populatePtp4lConf so there is no change to the ptpconfig_webhook.go file or to the code (previous solution before suggestion from Joseph to reuse the code)
Which one should we implement? Do we need a bugzilla to edit the ptpconfig_webhook.go even if no changes to existing code?

test/utils/testconfig/testconfig.go
edcdavid: yes, the original env DISCOVERY_MODE=true/false is replaced by PTP_TEST_MODE=OC/BC/DualNICBC/Discovery

test/utils/testconfig/testconfig.go
edcdavid: fixed

edcdavid: fixed

test/utils/testconfig/testconfig.go
aneeshkp: AS you are aware of cnf-test is one of the consumers of these tests. We need to retain the original env variable since this will break  cnf-test when we propagate these  changes to that project 
https://github.com/openshift-kni/cnf-features-deploy/blob/7a47e1d408013bc375abd8f8420c9a6488b88736/cnf-tests/testsuites/e2esuite/README.md

https://github.com/openshift-kni/cnf-features-deploy/blob/c5e4f1a70e472230c796921bfe86e068bd10c17a/cnf-tests/README.md#running-tests-in-discovery-mode


https://github.com/openshift-kni/cnf-features-deploy/blob/c5e4f1a70e472230c796921bfe86e068bd10c17a/cnf-tests/README.md#ptp-tests

edcdavid: fixed

test/utils/client/clients.go
```diff
@@ -85,3 +85,36 @@ func New(kubeconfig string) *ClientSet {
 
 	return clientSet
 }
+
+// GetTestClientSet Overwrites the existing clientholders with a mocked version for unit testing.
+func GetTestClientSet(k8sMockObjects []runtime.Object) *ClientSet {
```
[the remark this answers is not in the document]
aneeshkp: How do you run this test without a cluster? if KubeConfig is not set it will assume mock? What is the test path

josephdrichard:
Can you squash your changes into one commit?  There shouldn't be separate commits in the PR when it merges just to address comments.

api/v1/ptpconfig_webhook.go
aneeshkp: alright!, since  @josephdrichard has okayed to add these changes to webhook., let's keep it. No need to add bugsZilla since this code will be not be used by PTP.

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,295 @@
...
+	"context"
+
+	ptpv1 "github.com/openshift/ptp-operator/api/v1"
+	testclient "github.com/openshift/ptp-operator/test/utils/client"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ConfigCheck                  = "CONFIGCHECK"
```
aneeshkp, line 17: Do you mind adding comments to exported constants?

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	switch mode {
+	case OrdinaryClock, BoundaryClock, DualNICBoundaryClock, Discovery:
+		logrus.Infof("%s mode detected", mode)
+		GlobalConfig.PtpModeDesired = mode
+		GlobalConfig.Status = InitStatus
+		return GlobalConfig
+	case None:
+		logrus.Infof("No test mode specified using, %s mode. Specify the env variable PTP_TEST_MODE with one of %s, %s, %s, %s", OrdinaryClock, Discovery, OrdinaryClock, BoundaryClock, DualNICBoundaryClockString)
```
aneeshkp, line 152: The log should saying what is default config used when mode is none

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,295 @@
...
+	switch mode {
+	case OrdinaryClock:
+		return OrdinaryClockString
+	case BoundaryClock:
+		return BoundaryClockString
+	case DualNICBoundaryClock:
+		return DualNICBoundaryClockString
+	case Discovery:
+		return DiscoveryString
+	case None:
```
aneeshkp, line 98: Why None mode is required. What is None means,? If None is set then it should set to default which is an ordinary clock , right ?

edcdavid: right, none is just to support the case when the env variable is ""

test/utils/testconfig/testconfig.go
```diff
@@ -0,0 +1,295 @@
...
+	ConfiguredStatusString       = "configured"
+	DiscoverySuccessStatusString = "discoverySuccess"
+	DiscoveryFailureStatusString = "discoveryFailure"
+)
+
+type ConfigStatus int64
+
+const (
+	Start ConfigStatus = iota
+	InitStatus
```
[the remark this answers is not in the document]
edcdavid: fixed

test/utils/testconfig/testconfig.go
edcdavid: fixed

aneeshkp:
Running make functiontest on a SNO without ptp/master ,ptp/slave gives following error .
 need at least two nodes with ptp capable nics
      Expected
          <int>: 1
      to be >
          <int>: 1
          
>>  The test should discover its a single note

aneeshkp (approved):

PR #186, opened 2022-05-12
Upgrade Go to v1.18

sebrandon1:
Work in progress.  The one thing I can't figure out is where they are pulling the base image from:
registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6

This should also be supporting Go 1.18.


aneeshkp:
I think we are using 1.17 that's what openshift 4.11 is using.
https://github.com/openshift/ptp-operator/blob/master/Dockerfile.rhel7#L1

What we need is to upgrade must gather to 1.17 same as the main docker file That will be minimal changes and it is done using Bugzilla .

Also, this is not a testing tool requirement, if this is a bug we need to create a bugzilla.

aneeshkp:
openshift bot already has a PR, so we can close this PR
https://github.com/openshift/ptp-operator/pull/185

PR #187, opened 2022-05-12
Add shellcheck

sebrandon1:
Tests incoming PRs against shellcheck.  Watches the shell files in the `hack` folder.

Similar to: https://github.com/test-network-function/cnf-certification-test/pull/185

### DIFF
--- a/.github/workflows/pre-main.yaml
+++ b/.github/workflows/pre-main.yaml
@@ -1,0 +1,22 @@
+name: Test Incoming Changes
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/pre-main.yaml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-22.04
```
josephdrichard, line 13: Why Ubuntu and not rhel or fedora?

josephdrichard (commented):
Is there a bug this is linked to?

.github/workflows/pre-main.yaml
sebrandon1: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

PR #188, opened 2022-05-16
Update interface discovery process in multi node cluster

hamadise:


test/conformance/ptp/ptp.go
```diff
@@ -516,6 +516,18 @@ var _ = Describe("[ptp]", func() {
...
 func configurePTP() {
+	/*
+		The test takes significant amount of time checking
+		for the right interfaces to test
+		By setting the following variable environment for test purposes
+		The suite will check use them directly
+	*/
+
+	ptpGrandMasterNodeName := os.Getenv("PTP_CONFIG_MASTER_NODE")
+	masterInterfaceName := os.Getenv("PTP_CONFIG_MASTER_INTERFACE")
```
aneeshkp, line 527: what would be t default path if the environment is not set?

test/conformance/ptp/ptp.go
```diff
@@ -516,6 +516,18 @@ var _ = Describe("[ptp]", func() {
...
 
 func configurePTP() {
+	/*
+		The test takes significant amount of time checking
+		for the right interfaces to test
+		By setting the following variable environment for test purposes
+		The suite will check use them directly
+	*/
+
+	ptpGrandMasterNodeName := os.Getenv("PTP_CONFIG_MASTER_NODE")
```
aneeshkp, line 526: I feel this is not required. This should be native k8s way using by node label ptp/master and ptp/slave.

test/conformance/ptp/ptp.go
```diff
@@ -538,9 +568,14 @@ func configurePTP() {
 	if err == nil && configureFifo {
 		ptpSchedulingPolicy = "SCHED_FIFO"
 	}
-
 	for _, gmInterface := range ptpGrandMasterNode.InterfaceList {
+		if gmInterface != masterInterfaceName && masterInterfaceName != "" {
```
aneeshkp, line 572: There should be a separate validation to either fail if the desired (via env) interface is not in the master interface list/slave interface list  OR  if not found then pick up the default path and pick the default interface the way it did.

test/conformance/ptp/ptp.go
hamadise: an empty string.

test/conformance/ptp/ptp.go
hamadise: I did want the changes to be as minimal as possible. Hence, hijacking the current for loop. I'll add separate path and fall back to the regular discovery if it did fail

test/conformance/ptp/ptp.go
hamadise: By looking further in the code, we do label the nodes with ptp/master and ptp/slave after deciding which one is master and which one is slave.
what I can do, is assign master to a node with label node-role.kubernetes.io/control-plane or node-role.kubernetes.io/master; but this may not always be reflective of the clock role in the cluster.
I guess the gist of this is to provide developer with an easy way to specify which interfaces to use directly

aneeshkp:
please add a description to the PR  explaining in detail about this PR

test/conformance/ptp/ptp.go
aneeshkp: we will leave the node labeling logic as it is since it is expected in that way in the cnf-test repo too.
 If the user is going to specify the interface and he knows the interface nodes. then the user can always label the nodes as ptp/master and ptp/slave as required.

hamadise:
the PR is not necessary as we can preconfigure the cluster with PTP and test that configuration.

PR #189, opened 2022-05-16
Remove logr/klog, consolidate logging libraries

sebrandon1:
I noticed the the go.mod contained references to `logr`, `glog`, and `klog`.  `logr` was only being used in two places and `klog` was only being used in one place so I consolidated the calls for logging into `glog`.


controllers/ptpconfig_controller.go
```diff
@@ -47,8 +45,7 @@ type PtpConfigReconciler struct {
 //+kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
 
 func (r *PtpConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
-	reqLogger := r.Log.WithValues("Request.Namespace", req.Namespace, "Request.Name", req.Name)
-	reqLogger.Info("Reconciling PtpConfig")
+	glog.Infof("Request.Namespace %s Request.Name %s - Reconciling PtpConfig", req.Namespace, req.Name)
```
aneeshkp, line 48: The interface that Operator SDK (and controller-runtime) use for logging is actually logr, so any logging implementation that supports the logr interface can be **used.
These** are defaults used by operator-sdk.  By changing these there could be issues. We need to test these changes. Also think about  operator-sdk upgrades

pkg/leaderelection/leaderelection.go
```diff
@@ -31,7 +31,7 @@ func GetLeaderElectionConfig(restConfig *rest.Config, enabled bool) configv1.Lea
 				return leaderelection.LeaderElectionSNOConfig(defaultLeaderElection)
 			}
 		} else {
-			klog.Warningf("unable to get cluster infrastructure status, using HA cluster values for leader election: %v", err)
+			glog.Warningf("unable to get cluster infrastructure status, using HA cluster values for leader election: %v", err)
```
aneeshkp, line 34: Same here, klog implementation of logr. I believe all business logic , that is outside operator-sdk generated/managed code uses logr/zap/klog and the domain-related code is using glog.

aneeshkp:
again refactoring ptp operator code . Need better discussion and task added in detail.

controllers/ptpconfig_controller.go
sebrandon1: Interesting, this makes sense.  We can disregard this then.  Thanks!

sebrandon1:
Closing as not needed because of operator-sdk expecting certain libraries.

PR #190, opened 2022-05-16
Make context-use more consistent

sebrandon1:
context.TODO() and context.Background() are the same thing however they were not consistently being used.

Prior to this change, `Background()` had 60 uses and `TODO()` had 5 uses so I made every reference use context.Background().

### DIFF
--- a/main.go
+++ b/main.go
@@ -130,7 +130,7 @@ func createDefaultOperatorConfig(cfg *rest.Config) error {
 			DaemonNodeSelector: map[string]string{},
 		},
 	}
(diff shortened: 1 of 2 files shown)


aneeshkp:
These look like refactoring code. We shouldn't be looking  in to refactoring the main ptp operator code

aneeshkp:
@sebrandon1 . These become untraceable changes to the ptp-operator, we will have issues later if this produces a bug.
But  If we strongly believe that this has to be fixed within ptp operator,  then we need to create a Bugzilla and attach the Bugzilla to the PR. So it can be properly verified and traced.

PR #196, opened 2022-06-13
Reboot daemonset template

bnshr:
This template would be used by various PRs for PTP tests.


test/utils/daemonsets/daemonsets.go
```diff
@@ -0,0 +1,98 @@
...
+						},
+						{
+							Effect:            "NoExecute",
+							Key:               "node.kubernetes.io/unreachable",
+							Operator:          "Exists",
+							TolerationSeconds: &tolerationsSeconds,
+						},
+						{
+							Effect: "NoSchedule",
+							Key:    "node-role.kubernetes.io/master",
```
aneeshkp, line 80: Will this be an issue with the Single node, since the node will have a master label?

bnshr: I have not tested with SNO yet - I would get back once that is done. We have other issues in the test cluster and hence  there is a delay.

aneeshkp: you could remove that when SNO detected

aneeshkp (approved):

aneeshkp (approved):

aneeshkp:
/retitle Bug 2097810: PTP Test daemonset template

test/utils/daemonsets/daemonsets.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if !isReady {
+		return errors.New("daemonset debug pods not ready")
+	}
+
+	logrus.Infof("All the debug pods are ready.")
+	return nil
+}
+
+func isDaemonSetReady(status *appsv1.DaemonSetStatus) bool {
```
sebrandon1, line 215: The unit tests for this function could have been copied over as well: https://github.com/test-network-function/cnf-certification-test/blob/main/pkg/provider/provider_test.go#L91

PR #202, opened 2022-07-08
Implement PTP reboot-recovery test

bnshr:
This PR is to restart the slave node and verify that PTP is running after the reboot.
This test is configurable via the ENV variable `RUN_REBOOT_TEST`.
This includes also the information on how to run this test.


test/README.md
```diff
@@ -13,6 +14,11 @@ So for instance to run in discovery mode the command line could look like this:
 ```
 KUBECONFIG="/home/user/.kube/config" PTP_TEST_MODE=Discovery make functests
 ```
+To run all the tests
+```
+KUBECONFIG="/home/user/.kube/config" RUN_REBOOT_TEST= PTP_TEST_MODE=Discovery make functests
```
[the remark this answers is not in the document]
bnshr: Yes, it could be possible. Is there some test cases already for port_down?

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+
+	for _, recommend := range slavePtpConfig.Spec.Recommend {
+		matches := recommend.Match
+
+		for _, match := range matches {
+			nodeLabel := *match.NodeLabel
+			// Get all nodes with this node label
+			nodes, err := client.Client.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{LabelSelector: nodeLabel})
+			if err != nil {
+				panic(err.Error())
```
aneeshkp, line 1176: if you panic here , will it lose all test report?
why can't you fail reboot test in that case?

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+		containerName    = "container-00"
+		imageWithVersion = "quay.io/testnetworkfunction/debug-partner:latest"
+
+		timeout = 5 * time.Minute
+	)
+
+	// 1. create a daemon set for the node reboot
+	dsRunningPods, err := daemonsets.CreateDaemonSet(daemonSetName, defaultNamespace, containerName, imageWithVersion, timeout)
+	if err != nil {
+		logrus.Errorf("Error : +%v\n", err.Error())
```
aneeshkp, line 1156: Error to lowercase  `error : ` remove space too,

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+	checkSlaveSyncWithMaster(ptpRunningPods, fullConfig)
+
+	logrus.Info("Rebooting system ends ..............")
+}
+
+func rebootNode(pod v1core.Pod, node v1core.Node) {
+	checkRestart(pod)
+	// Wait for the node to be unreachable
+	const unrechable = false
+	testnode.WaitForNodeReachability(&node, 5*time.Minute, unrechable)
```
aneeshkp, line 1203: global constant 5*time.minute

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+
+	// Wait for all nodes to be reachable now after the restart
+	const reachable = true
```
aneeshkp, line 1206: try to use enum instead, 
type nodeStatus int
const (
	reachable nodeStatus = iota
        unreacble   
)

or have const clubbed together 
const (
reachable bool =true
unreacble bool= false
)

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+	testnode.WaitForNodeReachability(&node, 10*time.Minute, reachable)
+}
+
+func checkRestart(pod v1core.Pod) {
+	logrus.Info("Restarting a node")
+
+	const (
+		timeout         = 5 * time.Minute
```
aneeshkp, line 1214: This  timeout type of variable  is used many place , could it be moved globally

test/conformance/ptp/ptp.go
```diff
@@ -1081,3 +1138,237 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+
+	ptpSlaveRunningPods := []v1core.Pod{}
+	ptpMasterRunningPods := []v1core.Pod{}
+
+	for _, pod := range ptpPods.Items {
+		if podRole(pod, fullConfig.DiscoveredSlavePtpConfig.Label) {
+			waitUntilLogIsDetected(pod, 3*time.Minute, "Profile Name:")
+			ptpSlaveRunningPods = append(ptpSlaveRunningPods, pod)
+		} else if podRole(pod, fullConfig.DiscoveredMasterPtpConfig.Label) {
+			waitUntilLogIsDetected(pod, 3*time.Minute, "Profile Name:")
```
aneeshkp, line 1237: 3*time.Minute, same

test/conformance/ptp/ptp.go
```diff
@@ -31,10 +32,17 @@ import (
 	testclient "github.com/openshift/ptp-operator/test/utils/client"
 	"github.com/openshift/ptp-operator/test/utils/execute"
 	"github.com/openshift/ptp-operator/test/utils/nodes"
+	testnode "github.com/openshift/ptp-operator/test/utils/nodes"
 	"github.com/openshift/ptp-operator/test/utils/pods"
 	"github.com/openshift/ptp-operator/test/utils/testconfig"
 )
 
+const (
+	defaultNamespace        = "default"
```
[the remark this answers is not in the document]
bnshr: I am creating the reboot daemonset (unrelated to ptp config) in some already existing (in this case, default) namespace in the openshift cluster. And this DS is being deleted after the test. So, the namespace does not matter here, right?

test/conformance/ptp/ptp.go
```diff
@@ -76,7 +84,7 @@ var _ = Describe("[ptp]", func() {
 		It("Should check that all nodes are running at least one replica of linuxptp-daemon", func() {
 			By("Getting ptp operator config")
 
-			ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), "default", metav1.GetOptions{})
+			ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), defaultNamespace, metav1.GetOptions{})
```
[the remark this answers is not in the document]
bnshr: That is not a namespace, sadly!

test/conformance/ptp/ptp.go
```diff
@@ -992,3 +1055,248 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+	} else {
+		Expect(len(ptpSlaveRunningPods)).To(BeNumerically(">=", 1), "Fail to detect PTP slave pods on Cluster")
+	}
+	ptpRunningPods := append(ptpMasterRunningPods, ptpSlaveRunningPods...)
+	return ptpRunningPods
+}
+
+func verifyAfterRebootState(rebootedNodes []string, fullConfig testconfig.TestConfig) {
+	By("Getting ptp operator config")
+	ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), PtpLinuxDaemonNamespace, metav1.GetOptions{})
```
edcdavid, line 1166: name for ptpoperatoconfig should be "default"

test/conformance/ptp/ptp.go
```diff
@@ -866,7 +929,7 @@ func getNonPtpMasterSlaveAttachedInterfaces(pod *v1core.Pod) []string {
 }
 
 func enablePTPEvent() error {
-	ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), "default", metav1.GetOptions{})
+	ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), PtpLinuxDaemonNamespace, metav1.GetOptions{})
```
edcdavid, line 932: should use ptpConfigOperatorName for name

test/conformance/ptp/ptp.go
```diff
@@ -884,7 +947,7 @@ func enablePTPEvent() error {
 }
 
 func ptpEventEnabled() bool {
-	ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), "default", metav1.GetOptions{})
+	ptpConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(PtpLinuxDaemonNamespace).Get(context.Background(), PtpLinuxDaemonNamespace, metav1.GetOptions{})
```
edcdavid, line 950: should use ptpConfigOperatorName for name

bnshr: Right! Somehow my changes were ignored by my previous commit.

test/conformance/ptp/ptp.go
aneeshkp: WaitForNodeReachability the funx name does says what it is , so true or false value alters the behavior either waitForNodeReachbility  or not ?
Btw what is this trying to do ?
The un-reachable checks for  5 minutes  and  doesn't indicate if the node became un-reachable.
What if the node never restarted at this state and its still reachable?
Are you supposed to fail if the node never rebooted?

bnshr: WaitForNodeReachability is for waiting for the node reachability (reachable or unreachable) mentioned by the last argument. When you pass true, it means wait until the node is reachable. When you pass false, it means wait until the node is unreachable. Those variables are used for the readability of what this function is achieving for. 

In bare metal cluster, the timeout can be longer for the unreachability of the node, hence I plan to increase 5 mins to 10 mins. 

Are you supposed to fail if the node never rebooted?
- There is no way to check if the node is restarted within the timeout. Only way is to increase the timeout to very high and hope that the desired state is reached.

test/conformance/ptp/ptp.go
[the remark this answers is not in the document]
bnshr: @edcdavid Could you respond here the reason for using only default namespace here?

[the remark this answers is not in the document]
edcdavid: We are using the default namespace for the daemonset since the default namespace already has enough permissions to run the "oc debug" command. If not using the default namespace we'd have to configure privileged mode for this namespace when we otherwise would not need it ( e.g. oc adm policy add-scc-to-group privileged system:authenticated ).The daemonset is not related to ptp, it is just a way to access the platform

test/conformance/ptp/ptp.go
josephdrichard, line 932: Is this in "openshift-ptp" or "default"?

test/conformance/ptp/ptp.go
```diff
@@ -992,3 +1055,248 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
 	}
 }
+
+func rebootSlaveNode(fullConfig testconfig.TestConfig) {
+	logrus.Info("Rebooting system starts ..............")
+
+	const (
+		daemonSetName    = "ptp-reboot"
+		containerName    = "container-00"
+		imageWithVersion = "quay.io/testnetworkfunction/debug-partner:latest"
```
josephdrichard, line 1065: Is this image necessary here?  In my opinion qe should try to get required images in quay.io/openshift or openshift.io if at all possible, not elsewhere.

test/conformance/ptp/ptp.go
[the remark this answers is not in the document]
aneeshkp: ok this is not for PTP. Thanks

test/conformance/ptp/ptp.go
aneeshkp: @josephdrichard this is testframe pod used by cnf tests . currently it is at this repo. I don't think it will ever move  under opneshift. Since it is used only for e2e test , I think it is okay .

test/conformance/ptp/ptp.go
```diff
@@ -992,3 +1055,248 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+	const unrechable = false
+	testnode.WaitForNodeReachability(&node, timeoutIn5Minutes, unrechable)
+
+	// Wait for all nodes to be reachable now after the restart
+	const reachable = true
+	testnode.WaitForNodeReachability(&node, timeoutIn10Minutes, reachable)
+}
+
+func checkRestart(pod v1core.Pod) {
+	logrus.Info("Restarting a node")
```
edcdavid, line 1125: Add node name:
logrus.Info("Restarting a node: "+ pod.Spec.NodeName)

test/conformance/ptp/ptp.go
```diff
@@ -992,3 +1055,248 @@ func printInterface(config ptpv1.PtpConfig, interfaceDetailsMap map[string]strin
...
+	logrus.Info("Rebooting system starts ..............")
+
+	const (
+		daemonSetName    = "ptp-reboot"
+		containerName    = "container-00"
+		imageWithVersion = "quay.io/testnetworkfunction/debug-partner:latest"
+	)
+
+	// 1. create a daemon set for the node reboot
+	dsRunningPods, err := daemonsets.CreateDaemonSet(daemonSetName, PtpLinuxDaemonNamespace, containerName, imageWithVersion, timeoutIn5Minutes)
```
edcdavid, line 1069: Deploy the daemonset in default namespace

aneeshkp: Can we add better naming . ptp Linux daemonset and this is  confusing

test/conformance/ptp/ptp.go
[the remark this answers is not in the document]
bnshr: Interestingly, in `cnfdt13` cluster, we have permission to run the daemonset with privilege permission in `openshift-ptp` namespace, but in 3 nodes cluster not. 
For other clusters (if new are coming up), can we ensure that only `default` namespace would be having necessary permissions at service account level for the daemonset?

test/conformance/ptp/ptp.go
bnshr: good catch, it shall be default. Thanks!

test/conformance/ptp/ptp.go
bnshr: Sorry, I did not understand your comment. The namespace variable is changed, so we are not using PtpLinuxDaemonNamespace anymore. Is that what you suggested?

aneeshkp: we are using very generic name daemonset. There are two daemonset now here. ptpLinux daemonset and test harness daemonset . Hence the confusion

bnshr: @aneeshkp This is the reboot daemonset. I can change to rebootDaemonset.

aneeshkp (approved):

PR #209, opened 2022-08-09
http transport support to replace amq

aneeshkp:
add service to expose http transport endpoints
This addes http transport protocol for ptp events
Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>

### DIFF
--- a/bindata/linuxptp/ptp-daemon.yaml
+++ b/bindata/linuxptp/ptp-daemon.yaml
@@ -173,6 +173,28 @@ spec:
       targetPort: https
   sessionAffinity: None
   type: ClusterIP
+  {{ if (eq .EnableEventPublisher true) }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/scrape: "false"
+    service.beta.openshift.io/serving-cert-secret-name: linuxptp-daemon-secret
+  labels:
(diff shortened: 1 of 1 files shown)


bindata/linuxptp/ptp-daemon.yaml
```diff
@@ -173,6 +173,28 @@ spec:
...
+    app: linuxptp-daemon
+  name:  ptp-event-publisher-service
+  namespace: openshift-ptp
+spec:
+  selector:
+    app: linuxptp-daemon
+  clusterIP: None
+  ports:
+    - name: publisher-port
+      port: 9086
```
josephdrichard, line 194: Why port 9086?

josephdrichard (commented):
Is there a change to the bundle required?

bindata/linuxptp/ptp-daemon.yaml
aneeshkp: changed it to 9043, its the port that is running default publisher service (HTP  event Publisher service)
I just picked one that would conflict with the one most used

aneeshkp:
> Is there a change to the bundle required?

Not for this change I think . The code change is in side car and I am adding only new service to ptp operator which is deployed via  dindata/ folder. (Old method )  and not part of CSV

PR #212, opened 2022-08-22
Add HA option to ptpConfig

danielmellado:
This commit adds a new option to ptpConfig in order for the operator to
set up HA for clocks in the RU.

It partially implements the enhancement proposal at [1]

[1] https://github.com/openshift-kni/kni-enhancements/pull/4/

Signed-off-by: Daniel Mellado <dmellado@redhat.com>


aneeshkp:
is this changes targeted to 4.12?

api/v1/ptpconfig_types.go
```diff
@@ -61,11 +61,12 @@ type PtpConfigList struct {
...
-	Name        *string `json:"name"`
-	Interface   *string `json:"interface,omitempty"`
-	Ptp4lOpts   *string `json:"ptp4lOpts,omitempty"`
-	Phc2sysOpts *string `json:"phc2sysOpts,omitempty"`
-	Ptp4lConf   *string `json:"ptp4lConf,omitempty"`
+	Name            *string `json:"name"`
+	Interface       *string `json:"interface,omitempty"`
+	Ptp4lOpts       *string `json:"ptp4lOpts,omitempty"`
+	Phc2sysOpts     *string `json:"phc2sysOpts,omitempty"`
+	BoundaryClockHA *bool   `json:"boundaryClockHA,omitempty"`
```
aneeshkp, line 68: I think we should call it as EnableHA and the logic decided if config  has HA capability
It is not HA for boundary clock, it is more like HA between two PTP enabled nics  which could also apply to ordinary clock  right ?
May  be  if you create a Struct 
"HighAvailability": {
      "EnableHA": true|false,
      "PreferredPrimary": true|false,   // each nic will have initial state  request and status will  inform which nic is primary
      }
      Remember each nic has its own ptpConfig, in that case both has to have this section . 

     Probably best will be to introduce new  ptpHAConfig  and introduce controller to update HA preferences
monitor changes and update  linuxPtpDaemon of new change.
ptpHAConfig {
  "EnableHA": true|false,
  "HAPTPProfiles":
	- profile1
	- profile2 
  "PreferredPrimay": profile1
  "HeartBeatTimeOut": int
}

README.md
```diff
@@ -118,6 +119,7 @@ spec:
   - name: "profile1"
     ptp4lOpts: "-s -2"
     phc2sysOpts: "-a -r"
+    boundaryClockHA: false
```
aneeshkp, line 122: see comments below

api/v1/ptpconfig_types.go
danielmellado: +1 from my side. I'll modify this PR and rescope it so it matches the proposed change, and not only the field.

README.md
danielmellado: same

api/v1/ptpconfig_types.go
danielmellado: Would it make sense to have a HAPtPProfile? IIUC they would be just the same as the PTPProfiles, tbh. Anyway, I'll put my changes on the PtpConfig, let's iterate afterwards

danielmellado: Like, explaning myself, this is the sample for a multiple nic config:
```
apiVersion: ptp.openshift.io/v1
kind: PtpConfig
metadata:
  name: boundary-clock-ptpconfig
  namespace: ptp
spec:
  profile:
  - name: "profile1"
    ptp4lOpts: "-s -2"
    phc2sysOpts: "-a -r"
    ptp4lConf: |
      [ens7f0]
      masterOnly 0
      [ens7f1]
      masterOnly 1
      [ens7f2]
      masterOnly 1
  recommend:
  - profile: "profile1"
    priority: 4
    match:
    - nodeLabel: "node-role.kubernetes.io/worker"
```
I would just add the EnableHA and the different options there. Everything else should be the same, wdyt?

README.md
```diff
@@ -130,7 +161,12 @@ spec:
     priority: 4
     match:
     - nodeLabel: "node-role.kubernetes.io/worker"
+  highAvailability:
+    enableHA: true
+    preferredPrimary: "ens7f0"
```
aneeshkp, line 166: I still think HA config should  be outside ptpconfig , ptpConfig is per nic profile , so HA config new API  introduced , which  is read by linuxPtpdaemon to configure HA.
@josephdrichard

aneeshkp: also this requires to update clientset for ptp-opertaor

aneeshkp:
@danielmellado You  can do this in two phase
Phase 1 
1. Create simple configMap , which is static and the daemonset will look for this configmap to exists to enable HA
2.configmap example
apiVersion: v1
kind: ConfigMap
metadata:
  name: ptp-ha-configmap
  namespace: openshift-ptp
data:
   ha-policy-name: string
   primary-ptp-profile: profile1
   outofsync-wait-time : secs // time secs to wait HA enforced default 0
   outofsync-count: secs // how many times out of sync to enable HA
   max-offset-threshold:100 //optional to override ptpconfig values
   min-offset-threshold:-100 //optional to override ptpconfig values
   ha-enable-time: secs // after how many secs  after process starts enable HA poliocy
   retry-count: 0// after how many  failiover  time to abort
  3. The daemon can look for configmap and enable the logic
  Phase 2: 
  Once this is functioning we can add CRD and controller to create the config map

PR #228, opened 2022-09-14
bugfix in ptpconfig discovery,  fix for logging and additional logs

edcdavid:
Some bugfxes in ptpconfig discovery, logging, adding additional logs
also includes https://github.com/openshift/ptp-operator/pull/227 and skipping test old test waiting for clock ID when in discovery mode

### DIFF
--- a/test/conformance/ptp/ptp.go
+++ b/test/conformance/ptp/ptp.go
@@ -141,6 +141,12 @@ var _ = Describe("[ptp]", func() {
 		execute.BeforeAll(func() {
 			testconfig.CreatePtpConfigurations()
 			fullConfig = testconfig.GetFullDiscoveredConfig(utils.PtpLinuxDaemonNamespace, false)
+			if fullConfig.Status != testconfig.DiscoverySuccessStatus {
+				fmt.Printf(`ptpconfigs were not properly discovered, Check:
+- the ptpconfig has a %s label only in the recommend section (no node section)
+- the node running the clock under test is label with: %s`, utils.PtpClockUnderTestNodeLabel, utils.PtpClockUnderTestNodeLabel)
+				os.Exit(1)
+			}
 			if fullConfig.PtpModeDesired != testconfig.Discovery {
 				restartPtpDaemon()
 			}
@@ -1249,8 +1255,8 @@ func getProfileLogID(ptpConfigName string, label, nodeName *string) (id string, 
 	for _, pod := range ptpPods.Items {
 		isPodFound, err := pods.HasPodLabelOrNodeName(&pod, label, nodeName)
 		if err != nil {
-			logrus.Errorf("could not check %s pod role, err: %s", label, err)
-			Fail(fmt.Sprintf("could not check %s pod role, err: %s", label, err))
+			logrus.Errorf("could not check %s pod role, err: %s", *label, err)
+			Fail(fmt.Sprintf("could not check %s pod role, err: %s", *label, err))
 		}
 
 		if !isPodFound {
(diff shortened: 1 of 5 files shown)


test/utils/metrics/metrics.go
```diff
@@ -131,6 +131,12 @@ func CheckClockRoleAndOffset(ptpConfig *ptpv1.PtpConfig, label, nodeName *string
 	if nodeName == nil {
 		var name string
 		name, err = getNode(*label)
+		if err != nil || name == "" || label == nil || *label != utils.PtpClockUnderTestNodeLabel {
+			fmt.Printf(`error getting node name for label %s
```
aneeshkp, line 135: log instead of fmt

test/conformance/ptp/ptp.go
```diff
@@ -141,6 +141,12 @@ var _ = Describe("[ptp]", func() {
 		execute.BeforeAll(func() {
 			testconfig.CreatePtpConfigurations()
 			fullConfig = testconfig.GetFullDiscoveredConfig(utils.PtpLinuxDaemonNamespace, false)
+			if fullConfig.Status != testconfig.DiscoverySuccessStatus {
+				fmt.Printf(`ptpconfigs were not properly discovered, Check:
```
aneeshkp, line 145: log instead of fmt.

test/utils/metrics/metrics.go
```diff
@@ -90,9 +90,9 @@ func getMetric(nodeName, aIf, metricName string) (metric string, err error) {
 		var regex string
 		if metricName == OpenshiftPtpOffsetNs {
 			aIf = aIf[:len(aIf)-1] + "x"
-			regex = metricName + `{` + fromMaster + `iface="` + aIf + `",node="` + ptpPods.Items[index].Spec.NodeName + `",process="ptp4l"} ([0-9]*)`
+			regex = metricName + `{` + fromMaster + `iface="` + aIf + `",node="` + ptpPods.Items[index].Spec.NodeName + `",process="ptp4l"} (-*[0-9]*)`
 		} else {
-			regex = metricName + `{iface="` + aIf + `",node="` + ptpPods.Items[index].Spec.NodeName + `",process="ptp4l"} ([0-9]*)`
+			regex = metricName + `{iface="` + aIf + `",node="` + ptpPods.Items[index].Spec.NodeName + `",process="ptp4l"} (-*[0-9]*)`
```
aneeshkp, line 95: This should not be the case I think . There should be an metrics without from

test/utils/metrics/metrics.go
```diff
@@ -90,9 +90,9 @@ func getMetric(nodeName, aIf, metricName string) (metric string, err error) {
 		var regex string
 		if metricName == OpenshiftPtpOffsetNs {
 			aIf = aIf[:len(aIf)-1] + "x"
```
aneeshkp, line 92: could you use  regex for aIf, since its not significant

test/utils/metrics/metrics.go
```diff
@@ -131,6 +131,12 @@ func CheckClockRoleAndOffset(ptpConfig *ptpv1.PtpConfig, label, nodeName *string
 	if nodeName == nil {
 		var name string
 		name, err = getNode(*label)
+		if err != nil || name == "" || label == nil || *label != utils.PtpClockUnderTestNodeLabel {
```
aneeshkp, line 134: what would be the label incase of multi cluster for master(leader) and slave(follower) ? do you need to have that label defined

test/utils/metrics/metrics.go
edcdavid: This is the last log of the output, using fmt for better formatting

test/utils/metrics/metrics.go
edcdavid: This handles several metric types

PR #235, opened 2022-09-21
Ginkgo v2 parallel framework

bnshr:


hack/run-functests.sh
```diff
@@ -1,14 +1,15 @@
 #!/bin/bash
 set -x
 which ginkgo
+ginkgo version
```
josephdrichard, line 4: What are you checking for here?

hack/run-functests.sh
```diff
@@ -18,4 +19,13 @@ GOPATH="${GOPATH:-~/go}"
...
+VALIDATION_SUIT_SUBSTR="validation"
+
+if [[ $SUITE == *"$VALIDATION_SUIT_SUBSTR"* ]]
+then 
+	GOFLAGS=-mod=vendor ginkgo -v -p "$SUITE" --junit-report=$JUNIT_OUTPUT
+else 
+	# GOFLAGS=-mod=vendor ginkgo -v -p "$SUITE"/serial  "$SUITE"/parallel  --junit-report=$JUNIT_OUTPUT
+	GOFLAGS=-mod=vendor ginkgo -v -p "$SUITE"/parallel  --junit-report=$JUNIT_OUTPUT
+	# GOFLAGS=-mod=vendor ginkgo -v -p "$SUITE"/serial  --junit-report=$JUNIT_OUTPUT
+fi
```
josephdrichard, line 31: newline at end of file?

hack/run-functests.sh
bnshr: Removed this line which was introduced during debugging.

josephdrichard:
All yaml files in this project are named *.yaml not *.yml, so please follow that convention.

test/conformance/parallel/ptptestconfig.yml
```diff
@@ -0,0 +1,21 @@
...
+ failfast: true
+ minoffset: -100
+ maxoffset: 100
+ duration: 10
+testbcontinuous:
+ enable: true
+ failfast: true
+ minoffset: -100
+ maxoffset: 100
+ duration: 10
```
josephdrichard, line 21: missing new line at end of file.

josephdrichard (changes requested):

test/conformance/parallel/ptptestconfig.yml
bnshr: Added

bnshr:
> All yaml files in this project are named *.yaml not *.yml, so please follow that convention.

Changed to yaml in all places.

test/conformance/parallel/ptptestconfig.yaml
```diff
@@ -0,0 +1,21 @@
...
+ minoffset: -100
+ maxoffset: 100
+ duration: 10
+testacontinuous:
+ enable: true
+ failfast: true
+ minoffset: -100
+ maxoffset: 100
+ duration: 10
+testbcontinuous:
```
aneeshkp, line 16: won't this be duplicate ?Since you only have masteroffsetcontinuous 
we should only have that and not duplicate tests

test/conformance/parallel/ptptestconfig_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLoadConfiguration(t *testing.T) {
+	cfg := PtpTestConfig{}
+	path, _ := os.Getwd()
+	loadPtpTestConfig(fmt.Sprintf("%s/%s", path, "testdata/cfg1.yaml"), &cfg)
```
aneeshkp, line 14: how does this work when you use docker run ?
testdata/cfg1.yaml, will we have cfg2.yaml, cfg3.yaml , if yes, why ?
If there is one global configuration to override then we should have a one config with meaningful name for it .

test/conformance/serial/ptp.go
```diff
@@ -45,11 +46,6 @@ const (
 	masterOffsetLowerBound  = -100
```
aneeshkp, line 46: future changes , it will be good if we can configure this outside too , via env variable may before basic (default) test cases

test/README.md
```diff
@@ -34,6 +39,14 @@ for example:
 docker run -e PTP_TEST_MODE=OC -e ENABLE_TEST_CASE=reboot -v /home/usr/.kube/config.3nodes:/tmp/config:Z -v .:/output:Z quay.io/redhat-cne/ptp-operator-test:latest
 ```
 
+## Run SOAK test
+
+Soak tests are configurable through the test configuration file as YAML file stored at location `test/conformance/ptptestconfig.yaml`. These tests run in parallel.
```
aneeshkp, line 44: how do you override this configuration when you run vai docker run or makefile, can you pass the configuration file to override ?

test/conformance/parallel/ptptestconfig_test.go
bnshr: I think I did not understand your comments. How are other unit tests running from docker? `go test` from the right directory works.

bnshr: The config files inside the `testdata`directory are for unit tests. The config file that contains test configurations is ptpconfig.yaml

test/README.md
bnshr: I have the impression that the users of these tests would copy the entire codebase and change it manually and running Makefile or docker is not what they would use, please correct me here.

test/conformance/serial/ptp.go
bnshr: We can put these also under the global part of the test config, no?

test/conformance/parallel/ptptestconfig.yaml
bnshr: There is a plan to put these variables as global inside the config.

PR #245, opened 2022-09-30
Fix skipRange with double quotes

jzding:
When ART builds the bundle image the skipRange regexp should exactly
match expected regexp in art.yaml. Currently the upper bound of skipRange
is not auto-updating because we use single quote in the skipRange string.

Signed-off-by: Jack Ding <jackding@gmail.com>


aneeshkp (approved):

nishant-parekh (approved):

jzding:
cherry-pick release-4.11

nishant-parekh:
as the range is different not sure if this can be cherrypicked..

PR #249, opened 2022-10-05
Bug OCPBUGS-2042: Fix skipRange with double quotes

jzding:
When ART builds the bundle image the skipRange regexp should exactly match expected regexp in art.yaml. Currently the upper bound of skipRange is not auto-updating because we use single quote in the skipRange string.

Signed-off-by: Jack Ding <jackding@gmail.com>


josephdrichard:
What happens when you run make bundle?  Does it overwrite these changes?

nishant-parekh:
> 

make bundle only recreates the bundle directory. The changes to config directory will persist and bundle directory is created again using the config directory file, so it would persist as well

jzding:
> What happens when you run make bundle? Does it overwrite these changes?

When I run `make bundle` it overwrites double quotes with single quotes so I have to change them back.

josephdrichard (changes requested):
The bundle should just be the output of 'make bundle'.  Whenever a designer makes a change that affects the manifest or somewhere related, it is expected that they'll just run make bundle to regenerate it.  We can't expect that they'll just manually make this change every time as well, or I expect it would eventually get merged where someone forgets to do this.  Can you find why make bundle has different output then required?  If this can't be fixed with a change in the kutomize command, then worst case scenario, this should be done with sed inside of the makefile.  We shouldn't ever manually be modifying the bundle.

jzding:
> The bundle should just be the output of 'make bundle'. Whenever a designer makes a change that affects the manifest or somewhere related, it is expected that they'll just run make bundle to regenerate it. We can't expect that they'll just manually make this change every time as well, or I expect it would eventually get merged where someone forgets to do this. Can you find why make bundle has different output then required? If this can't be fixed with a change in the kutomize command, then worst case scenario, this should be done with sed inside of the makefile. We shouldn't ever manually be modifying the bundle.

I have fixed `make generate & make manifests` and `make bundle`.

nishant-parekh (approved):

PR #253, opened 2022-10-13
Proposal for temporary ginkgo v1 custom reporter

greyerof:
In order for TCs to report extra information as with ginkgo v2's AddReportEntry function, I've made a backport of such function into ginkgo v1, so it can be used (if needed) by ptp test cases until migration to v2 is done. I've copied the signature so when v2 is used, the changes should be minimal.

Whatever data added with AddReportEntry will appear always in the system-out field of the corresponding TC, no matter whether the TC failed or passed.

To use it inside any TC, just import the package and use it:

```
import ptpReporter
"github.com/openshift/ptp-operator/test/utils/ginkgo_reporter"

It(.. {
  ...
  ptpReporter.AddReportEntry("myEntry1", "this is the text for my entry")
  ...
})
```
The xml output for that TC will be:
```
<testcase name="..." ...>
  <system-out>Report Entries:&#xA;myEntry1&#xA;this is the text for my&#x9;entry&#xA;</system-out>
</testcase>
```
To mimic the "AddReportEntry" function in v1, I created a custom ginkgo reporter for ptp, which is a decorator of the ginkgo's JUnitReporter that overrides the function that is called by ginkgo when all the test cases have finished. This function is responsible for creating the xml JUnit file.

To save the entries for each TC, the PTP ginkgo reporter needs to be saved as a "singleton" package variable, so the entries are kept until the xml is generated. The ptp custom reporter calls the underlying junit reporter to create the xml file. Then, it will be read and reparsed so all the stored report entries can be added to the corresponding test cases. Finally, the xml file is regenerated with this new information appearing in the <system-out> xml nodes.

As in ginkgo v2, any object can be passed as the entry value, as it will be serialized using the stringer function (if available) or with fmt.Sprintf("%+v", value) as fallback method.

When switching to ginkgo v2, the standard JUnitReporter can be used again, and the prefix "ptpReporter." (or whatever import alias) must be removed from all the calls to AddReportEntry().

The UTs file seems large, but it has two simple parts:
  1. Custom UTs (using ginkgo) for checking the AddReportEntry functionality: makes sure the system-out xml node is updated accordingly.
  2. Copy/pasted UTs from ginkgo v1's junit_reporter_test.go to make sure everything works the same if no report entries are used in any test cases (regression tests).


test/utils/ginkgo_reporter/ptp_reporter_test.go
```diff
@@ -0,0 +1,633 @@
+package ptp_reporter
+
+import (
+	"encoding/xml"
+	"io/ioutil"
```
aneeshkp, line 5: "io/ioutil" has been deprecated

greyerof: Fixed.

josephdrichard:
This will be reverted as soon as ginkov2 merges?

greyerof:
@josephdrichard yes, that's the idea. When ginkgo v2 is used, this can be reverted, but the calls to this backported "AddReportEntry" function should remain, but without the package prefix.

With ginkgo v1 + this PR:
```
ptpReporter.AddReportEntry("myEntry", "myEntry data string")
```
With ginkg v2:
```
AddReportEntry("myEntry", "myEntry data string1")
```

test/conformance/test_suite_test.go
```diff
@@ -42,7 +43,7 @@ func TestTest(t *testing.T) {
 
 	rr := []Reporter{}
 	if junitPath != nil {
-		rr = append(rr, reporters.NewJUnitReporter(*junitPath))
+		rr = append(rr, ptpReporter.NewPTPJUnitReporter(*junitPath))
```
josephdrichard, line 46: Can you add a TODO comment about the change here when this is upgraded to ginkov2?

greyerof: Done.

josephdrichard:
/retest
/lgtm
@bnshr please take a look before this merges, as you're working on Gingko v2.

PR #282, opened 2022-12-09
Adding multi-node support for ptp events

edcdavid:
Work in progress!!

This PR creates one  ptp-event-publisher-service-xxx per node and an ingress object per cluster for ptp-events. For the ingress object, one backend is created per node to direclty access the ptp events service on that node.
image
requires an new nodeName label on the linux-ptp-daemon. See related PR: https://github.com/openshift/linuxptp-daemon/pull/107

### DIFF
--- a/bindata/linuxptp/event-service.yaml
+++ b/bindata/linuxptp/event-service.yaml
@@ -1,0 +1,20 @@
+---
+apiVersion: v1
+kind: Service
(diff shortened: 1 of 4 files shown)


aneeshkp (approved):

aneeshkp (approved):
/lgtm

PR #286, opened 2022-12-19
CPU utilization test case for the parallel test suite.

greyerof:
The cpu usage is measured once per minute until the TC's configured timeout is reached or the failure threshold is reached, whatever comes first. Every cpu check cycle, the cpu usage is retrieved for each ptp-daemon and ptp-operator pods. For each node, the sum of cpu usage of both deployed pods (ptp-operator pod + ptp daemonset pod) is compared against the hardcoded threshold.

To get the cpu usage, the value for the "container_cpu_usage_seconds_total" prometheus metric is retrieved. This is just a cumulative counter that prometheus gets every 30 secs from cAdvisor, so, the "rate" function needs to be used. Prometheus' `rate()` function gets the first and the last sample in the selected vector (last 70secs samples, 2 samples in this case), subtracts them, and divides the result by the requested time window (70secs). It is worth noting that [prometheus might do some extrapolations](https://promlabs.com/blog/2021/01/29/how-exactly-does-promql-calculate-rates).

This way to get the cpu usage "improves" the detection of cpu spikes: now, the obtained cpu usage is a rate for the last 70 seconds, while the previous implementation was obtaining an averaged value for the last 5 minutes (using PodMetrics resource). The downside of this is that, if the scrape interval is modified/updated to be higher than 30secs, the `rate` function won't have enough samples on the requested interval (70secs), and it won't return any value.

Future improvements: 
- Make the cpu usage threshold a configurable param via yaml file.
- Get the configured scrape interval to make sure it's <= 30 secs. Making the time window for `rate` a configurable param could help on this, too.

A different approach would be not using prometheus at all, and get the pods cpu usage directly from each of the nodes using privilege daemonset's pods. We could get the cgroups'  `cpuacct.usage` on a per-second basis. This approach involves getting the cgroup ID of each pod on each node and cat'ing the /sys/fs/cgroup/.../cpuacct/cpuacct.usage value. It's a lot of extra work, so I'm not sure this is a good idea for this test case purpose...


josephdrichard (commented):
Is this intended to just take a snapshot of the usage at a specific time, or will it detect spikes?

test/conformance/parallel/ptp.go
```diff
@@ -113,6 +116,78 @@ func testPtpSlaveClockSync(fullConfig testconfig.TestConfig, testParameters ptpt
...
+	const (
+		minimumFailureThreshold  = 1
+		cpuUsageCheckingInterval = 1 * time.Minute
+		milliCoresThreshold      = ptptestconfig.PtpDefaultMilliCoresUsageThreshold
+	)
+
+	logrus.Debugf("CPU Utilization TC Config: %+v", testParameters.SoakTestConfig.CpuUtilization)
+
+	params := testParameters.SoakTestConfig.CpuUtilization.TestSpec
+	if !params.Enable {
```
bnshr, line 133: Here we need to check also if all soak tests are disabled or not.

test/conformance/parallel/ptp.go
```diff
@@ -113,6 +116,78 @@ func testPtpSlaveClockSync(fullConfig testconfig.TestConfig, testParameters ptpt
...
+	// Create ticker for cpu usage checker function.
+	cpuUsageCheckTicker := time.NewTicker(cpuUsageCheckingInterval)
+
+	// Set failureThresold limit number.
+	failureThreshold := minimumFailureThreshold
+	if params.FailureThreshold > minimumFailureThreshold {
+		failureThreshold = params.FailureThreshold
+	}
+
+	logrus.Infof("Running TC for %s (failure threshold: %d)", testCaseDuration.String(), failureThreshold)
```
bnshr, line 151: Could we replace TC with test please? It may mean anything total compensation, terminal condition, transfer certificate etc etc. ;)

greyerof:
@josephdrichard It relies on the PodMetrics resource from the metrics-server, which gives an average cpu usage in the (preconfigured) time window, which seems to be 5mins. I'm getting that resource in a per-minute basis. That's the simplest way I found to get the cpu usage. For detecting shorter cpu usage spikes, I need to find a way to (a) decrease the time window for PodMetrics measures or (b) measure the cpu manually every "N" (1,2,5...) seconds. Option (b) requires more investigation. Any suggestion here?

test/conformance/parallel/ptp.go
greyerof: Done.

test/conformance/parallel/ptp.go
greyerof: Done.

bnshr (approved):

aneeshkp (commented):
when event is enabled you will have another sidecar , can you detect that and include int his test case?

greyerof:
> when event is enabled you will have another sidecar , can you detect that and include int his test case?

@aneeshkp can't confirm yet, but I think the current code can also handle it. As it's just another container in the pod, its cpu usage should appear in the corresponding PodMetrics resource, so the current implementation is already prepared to accumulate it to get the total cpu usage for that pod. See file [test/pkg/ptptesthelper/ptptesthelper.go](https://github.com/openshift/ptp-operator/pull/286/files#diff-00124ea6f95e427da2c926fc9a9bb45dce3d2d01faf81cbb5e95c79c4e749354), function `GetPodTotalCpuUsage`.

test/conformance/parallel/ptp.go
```diff
@@ -113,6 +116,102 @@ func testPtpSlaveClockSync(fullConfig testconfig.TestConfig, testParameters ptpt
...
+		failureThreshold = params.FailureThreshold
+	}
+
+	prometheusPod, err := ptptesthelper.GetPrometheusPod()
+	Expect(err).To(BeNil(), "failed to get prometheus pod")
+
+	ptpPodsPerNode, err := ptptesthelper.GetPtpPodsPerNode()
+	Expect(err).To(BeNil(), "failed to get ptp pods per node")
+
+	// White until prometheus can scrape a couple of cpu samples from ptp pods.
```
edcdavid, line 155: typo White -> Wait

test/pkg/ptptesthelper/ptptesthelper.go
```diff
@@ -322,3 +322,100 @@ func RebootSlaveNode(fullConfig testconfig.TestConfig) {
...
+// to return the total cpu usage by all the given pods.
+// As each query needs to be done inside one of the prometheus pods, an optional
+// param prometheusPod can be set for that purpose. If it's nil, the function
+// will try to get it on every call.
+func GetPodsTotalCpuUsage(pods []*corev1.Pod, prometheusPod *corev1.Pod) (float64, error) {
+	const (
+		// To make sure that prometheus can use rate() with at least two samples,
+		// we should use at least two sampling periods (2 * 30s = 60). We'll add 10
+		// extra seconds as a safeguard.
+		timeWindow = 70 * time.Second
```
edcdavid, line 372: improvement only(not necesarily in this PR): read scrape_interval from the prometheus container and dynamically adjust the time window.

aneeshkp: Can we capture these improvement as a issue or task so we don't lose all these good points?

test/conformance/config/ptptestconfig.go
```diff
@@ -11,6 +11,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const (
+	// Default cpu usage threshold in milliCpus.
+	PtpDefaultMilliCoresUsageThreshold = 15
```
aneeshkp, line 16: In real cluster I saw ptp daemon crossing avg 17 mc . So this could fail  most of the time. Also cpu usage changes for different profile OC/BC and Dual as ptp process increases .

aneeshkp (approved):
/lgtm

PR #288, opened 2023-01-08
Validation tests: Check pods status in an Eventually

sabinaaledort:
When running the validation tests just after the operator deployment the tests can fail if the pods are not ready yet. To avoid this, the pods validation should be done in an Eventually.

### DIFF
--- a/test/validation/tests/validation.go
+++ b/test/validation/tests/validation.go
@@ -14,6 +14,7 @@ import (
 
 	testutils "github.com/openshift/ptp-operator/test/pkg"
 	testclient "github.com/openshift/ptp-operator/test/pkg/client"
+	"github.com/openshift/ptp-operator/test/pkg/ptphelper"
 )
 
 var _ = Describe("validation", func() {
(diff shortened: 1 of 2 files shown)


test/validation/tests/validation.go
```diff
@@ -24,22 +24,57 @@ var _ = Describe("validation", func() {
...
+
+				return nil
+			}, testutils.TimeoutIn10Minutes, testutils.TimeoutInterval2Seconds).ShouldNot(HaveOccurred())
 		})
 
 		It("should have the linuxptp daemonset in running state", func() {
-			daemonset, err := testclient.Client.DaemonSets(testutils.PtpNamespace).Get(context.Background(), testutils.PtpDaemonsetName, metav1.GetOptions{})
-			Expect(err).ToNot(HaveOccurred())
-			Expect(daemonset.Status.NumberReady).To(Equal(daemonset.Status.DesiredNumberScheduled))
+			Eventually(func() error {
```
edcdavid, line 54: What about using the existing WaitForPtpDaemonToBeReady function here?

sabinaaledort: I was not familiar with it. Thanks! Fixed.

edcdavid (approved):

aneeshkp (approved):

PR #290, opened 2023-01-10
Continuous clock synchronization tests using events

edcdavid:
This is to add support for test cases driven by the cloud-event-proxy framework. The first of such test cases is monitoring clock synchronization continuously and failing if the number of loss of sync exceeds a configured threshold. The event collected during the test run are saved into a CSV file.
This PR uses the following new projects (in redhat-cne):
- https://github.com/redhat-cne/ptp-listener-lib : this is the event listener library to remotely connect to the  ptp-operator cloud-event-proxy  
- https://github.com/redhat-cne/ptp-listener-exports: holds common types
- https://github.com/redhat-cne/channel-pubsub: publisher-Subscriber library relying on native Golang channels. This is used to propagate events towards the test that subscribe for them.

The following pre-existing project provides port forwarding to the ptp-daemon-pods:
- https://github.com/anthhub/forwarder
Also disabled phc2sys synchronization in the grandmaster to improve stability. This is because in certain tests downstream oc and grandmaster can share the same node. the clock under test always sync up its os clock

When running parallel test cases, ginkgo actually runs test cases on different processes (not different threads). Events are propagated in the following way:
- for a given test case, 2 ports are allocated, one for the localhost tunnel endpoint, (8901 for syncState testcase) and another for the local http server listening for events post (8902 for syncState testcase).
- within a test case (e.g. within a process) events are dispatched with a channel-pubsub internally.
- any test cases can register any combination of events  

There are 2 modes to register events using (USE_PTP_EVENT_CONSUMER_SIDECAR env variable):
- using a consumer sidecar: this is the official way
- using the publisher cloud-event-proxy container in the ptp pods 

Also contains:
- a fix for the uds test regression
- an environment variable to set the log level (PTP_LOG_LEVEL)
- This PR also refactor the pod logging mechanism, now using streaming logs instead of repetitively downloading the logs. Regex are also used across the board to ease parsing and offer better error logs. 
- refactor to use the global config in all test cases


test/pkg/testconfig/testconfig.go
```diff
@@ -51,7 +52,7 @@ const (
 	DualNICBoundaryClockString = "DualNICBC"
 	ptp4lEthernet              = "-2"
 	ptp4lEthernetSlave         = "-2 -s"
-	phc2sysGM                  = "-a -r -r -n 24"
+	phc2sysGM                  = "" // do not autoconfigure phc2sys, use local IF clock as clock source for grandmaster
 	phc2sysSlave               = "-a -r -n 24"
```
aneeshkp, line 56: for events you need 
ptp4lOpts: "-2 --summary_interval -4" 
phc2sysOpts: "-a -r -m -n 24 -N 8 -R 16"

edcdavid: added the new parameters

test/conformance/parallel/consumer-sidecar.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: cloud-proxy-sidecar
+  namespace: openshift-ptp
```
aneeshkp, line 5: could you deploy this to test namespace? Its easy to clean on failures

test/conformance/parallel/consumer-sidecar.yaml
```diff
@@ -0,0 +1,53 @@
...
+  labels:
+    app.kubernetes.io/name: proxy
+spec:
+  containers:
+  - name: sidecar
+    image: quay.io/redhat-cne/cloud-event-proxy
```
aneeshkp, line 11: Please use image form ptp-operator . rehdat-cne image may not match with version that is deployed by ptp operator

test/conformance/parallel/consumer-sidecar.yaml
```diff
@@ -0,0 +1,53 @@
...
+    args:
+            - "--metrics-addr=127.0.0.1:9091"
+            - "--store-path=/store"
+            - "--transport-host=consumer-events-subscription-service.openshift-ptp.svc.cluster.local:9043"
+            - "--http-event-publishers=ptp-event-publisher-service-master0.openshift-ptp.svc.cluster.local:9043"
```
aneeshkp, line 16: instead of master0,  you  can use NODE_NAME, it will be replaced by referencing to nodename that pod is running on . Again this is only > 4.13  and master support .this  will not work  for any release below 4.13. I sit possible to add that check and skip this tests

aneeshkp: we might backpor to 4.12 http support

test/conformance/parallel/consumer-sidecar.yaml
```diff
@@ -0,0 +1,53 @@
...
+        containerPort: 9043
+  volumes:
+    - name: pubsubstore
+      emptyDir: {}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: consumer-events-subscription-service
+  namespace: openshift-ptp
```
aneeshkp, line 43: have them under their own namespace

test/conformance/parallel/consumer-sidecar.yaml
edcdavid: This file is not part of the feature, just a manual helper. Will remove the file

test/conformance/parallel/consumer-sidecar.yaml
edcdavid: fixed

test/conformance/parallel/consumer-sidecar.yaml
edcdavid: fixed

test/conformance/parallel/parallel_suite_test.go
```diff
@@ -45,17 +51,52 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	err := testconfig.CreatePtpConfigurations()
 	if err != nil {
 		Fail(fmt.Sprintf("Could not create a ptp config, err=%s", err))
 	}
-
+	By("Refreshing configuration", func() {
+		ptphelper.WaitForPtpDaemonToBeReady()
+		fullConfig = testconfig.GetFullDiscoveredConfig(pkg.PtpLinuxDaemonNamespace, true)
+	})
 	ptphelper.RestartPTPDaemon()
```
aneeshkp, line 63: will this restart Daemon affect any other parallel tests for an example CPU usage etc 
or is this happens at the beginning of test case ?

test/pkg/event/event.go
```diff
@@ -1,12 +1,305 @@
...
+					Name: "pubsubstore",
+					VolumeSource: corev1.VolumeSource{
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+					},
+				},
+			},
+			ServiceAccountName: TestSidecarSaName,
+			Containers: []corev1.Container{
+				{Name: TestSidecarContainerName,
+					Image: "quay.io/redhat-cne/cloud-event-proxy",
```
aneeshkp, line 69: can you use quay.io/openshift/cloud-event-proxy

test/pkg/event/event.go
```diff
@@ -1,12 +1,305 @@
...
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: pointer.Bool(true),
+						RunAsUser:  rootUser,
+					},
+
+					Args: []string{"--metrics-addr=127.0.0.1:9091",
+						"--store-path=/store",
+						"--transport-host=consumer-events-subscription-service." + ConsumerSidecarTestNamespace + ".svc.cluster.local:9043",
+						"--http-event-publishers=ptp-event-publisher-service-" + nodeName + "." + ProviderSidecarTestNamespace + ".svc.cluster.local:9043",
```
aneeshkp, line 78: is this full node name or just the first part?

test/conformance/config/ptptestconfig.yaml
```diff
@@ -4,6 +4,7 @@ global:
  minoffset: -100
 soaktest:
   disable_all: false
+  EventOutputFile: "./event-output.csv"
```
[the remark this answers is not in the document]
edcdavid: fixed

test/pkg/event/event.go
edcdavid: fixed. Now getting the cloud-events-proxy image from the linux-ptp-daemon pods

test/pkg/event/event.go
edcdavid: Fixed. Thanks, good catch!

aneeshkp (approved):

PR #305, opened 2023-02-15
Bash script to get the cpu usage of ptp containers.

greyerof:
Simple bash tool for developers to get the cpu usage of PTP containers. This tool is not intented for use in any production/release system. It needs a previously exported KUBECONFIG env var.

This script polls an openshift cluster with the "curl" tool to get the cpu usage of each PTP container found in a cluster. The cpu usage comes from the prometheus metric container_cpu_usage_total_seconds, using the rate() function on samples from the last N seconds. This time window is configurable via env var. Each container cpu usage value will be sent to stdout and to a csv file (default ptp-cpu-usage.csv) which is removed at the begining in case it already exists.

### DIFF
--- a/test/scripts/ptp-cpu-usage.sh
+++ b/test/scripts/ptp-cpu-usage.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+#
+# Simple bash tool for developers to get the cpu usage of PTP containers. This tool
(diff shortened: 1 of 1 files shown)


aneeshkp (approved):
/lgtm

PR #306, opened 2023-02-15
BUG OCPBUGS-7441: Match the O-RAN specification for the ocloudNotifications URI

mrVectorz:
Adds the o in front of cloudNotifications URI per O-RAN specification.
BUG OCPBUGS-7441


aneeshkp (approved):
/lgtm

PR #310, opened 2023-02-17
[release 4.12] OCPBUGS-7728: Fix amq race condition failing on service check and repeating same check on every reconcile

aneeshkp:
Fix amq race condition failing on service check and repeating same check on every reconcile


jzding (approved):
/lgtm

PR #313, opened 2023-02-22
Enable e810 plugin by default

josephdrichard:


controllers/ptpoperatorconfig_controller.go
```diff
@@ -214,7 +214,15 @@ func (r *PtpOperatorConfigReconciler) syncLinuxptpDaemon(ctx context.Context, de
 		}
 	}
 
-	enabledPlugins := strings.Join(defaultCfg.Spec.EnabledPlugins, ",")
+	var pluginList []string
+
+	if defaultCfg.Spec.EnabledPlugins != nil {
+		pluginList = *defaultCfg.Spec.EnabledPlugins
+	} else {
+		pluginList = []string{"e810"} // Enable e810 by default if plugins not specified
```
aneeshkp, line 222: can we have e810 as  map  in api types, so in future usecase ptpdaemon can reference from there?
[string]string{}

controllers/ptpoperatorconfig_controller.go
```diff
@@ -214,7 +214,15 @@ func (r *PtpOperatorConfigReconciler) syncLinuxptpDaemon(ctx context.Context, de
...
+	}
+
+	enabledPlugins := strings.Join(pluginList, ",")
 	data.Data["EnabledPlugins"] = enabledPlugins
 	if enabledPlugins != "" {
```
aneeshkp, line 227: So the logic is you control plugins enable and disable options from ptpoperator ,correct ?

josephdrichard: This would set the environment variable the daemon uses to read which plugins to register, and then the plugin will act based on the ptpconfig.
We can revisit this in the future when we are working on hwconfig population from the plugin.

aneeshkp: understood

bundle/manifests/ptp.openshift.io_ptpoperatorconfigs.yaml
```diff
@@ -44,9 +44,9 @@ spec:
                   type: string
                 type: object
               plugins:
-                items:
-                  type: string
-                type: array
+                additionalProperties:
+                  x-kubernetes-preserve-unknown-fields: true
```
aneeshkp, line 48: I am not sure what implication will have in future upgrades or using conversion webhook.
Is this safe to use ? any implications ?

controllers/ptpoperatorconfig_controller.go
aneeshkp: Think of operator upgrades when you design to depreciate any variables

manifests/stable/ptp.openshift.io_ptpoperatorconfigs.yaml
```diff
@@ -44,9 +44,9 @@ spec:
                   type: string
                 type: object
               plugins:
-                items:
-                  type: string
-                type: array
+                additionalProperties:
+                  x-kubernetes-preserve-unknown-fields: true
```
aneeshkp, line 48: Same here , need to check  if this will be an issue if we use conversion webhook.

aneeshkp: https://github.com/operator-framework/operator-lifecycle-manager/blob/c652154fd7eeb72a23f78856b2cbade775efad41/pkg/controller/install/webhook.go#L234

Looks like it has to be set to false for conversion webhook to work.

PR #316, opened 2023-03-02
fix for junit file not recorded and updated README. Log fixes.

edcdavid:
This is a fix for a regression with the junit output file not generated. Also added some log fixes


aneeshkp (approved):
/lgtm

PR #331, opened 2023-03-30
[release-4.12] Fix URI for PTP events conformance tests

javierpena:
[1] changed the URI for the cloud notifications URI to be ocloudNotifications instead of cloudNotifications. This patch updates the PTP events conformance test to match.

[1] - https://github.com/redhat-cne/cloud-event-proxy/commit/e527530f613ff0d5cecadccdeec3a24ff1be913b


jzding (approved):
/lgtm

PR #348, opened 2023-05-17
Add k8s reporter and junit reporter

liornoy:
Since we're working on splitting the cnf-tests, we must keep producing the same reports for
failed test cases with a k8s reporter, and a custom junit report.

This PR modifies the serial test suite that is invoked by the cnf-tests and adds those reporters.


liornoy:
Closing this one because we are thinking of a new direction, using the default junit reporter, but just processing its output differently.

PR #353, opened 2023-05-31
Add k8s reporter to e2e tests

liornoy:
This PR utilizes the openshift-kni k8sreporter, and passes it the CRDs, APIs and namespace of ptp.
The reporter will function only if the `-report` flag is passed.

This is to be aligned with the current reporter we use in the cnf-tests.

Currently created as a draft until the official proposal is filed and reviewed.


liornoy:
closing this for now, because I'm not sure if we still going to test PTP as part of the cnf-tests.

PR #358, opened 2023-06-19
OCPBUGS-14993: update operator bundle version

vitus133:
This commit updates operator version to 4.14


vitus133:
Apparently there is already a PR for this

PR #359, opened 2023-06-24
CNF-7015: Add basic unit test for plugin framework

josephdrichard:
Enables generic plugin, verifies that expect log is present, then disables it again.
Further commits will add support for testing the APIs

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -26,5 +26,5 @@ if [[ $SUITE == *"$VALIDATION_SUIT_SUBSTR"* ]]
 then
 	GOFLAGS=-mod=vendor ginkgo --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"
 else
-	GOFLAGS=-mod=vendor ginkgo -keepGoing --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial "$SUITE"/parallel
+	GOFLAGS=-mod=vendor ginkgo -keepGoing --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial
 fi
--- a/test/conformance/serial/ptp.go
+++ b/test/conformance/serial/ptp.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"sort"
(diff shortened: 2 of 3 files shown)


test/conformance/serial/ptp.go
```diff
@@ -533,6 +533,43 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+				foundMatch := false
+				for podIndex := range ptpPods.Items {
+					_, err := pods.GetPodLogsRegex(ptpPods.Items[podIndex].Namespace,
+						ptpPods.Items[podIndex].Name, pkg.PtpContainerName,
+						pluginLog, true, pkg.TimeoutIn3Minutes)
+					if err != nil {
+						logrus.Errorf(fmt.Sprintf("Reference plugin not loaded, err=%s", err))
+						continue
+					}
+					foundMatch = true
```
jzding, line 568: I don't think GetPodLogsRegex returning err==nil means a match is found.
Could you fix the function GetPodLogsRegex()?
https://github.com/openshift/ptp-operator/blob/master/test/pkg/pods/pods.go#L182
This last line should be
```
return matches, fmt.Errorf(""no matches found")
```

josephdrichard: It will return this error if there's no match:
https://github.com/openshift/ptp-operator/blob/master/test/pkg/pods/pods.go#L161

Don't want to make any changes there as part of this PR so I don't break anything else; it can be done in a followup PR if necessary

hack/run-functests.sh
```diff
@@ -26,5 +26,5 @@ if [[ $SUITE == *"$VALIDATION_SUIT_SUBSTR"* ]]
 then
 	GOFLAGS=-mod=vendor ginkgo --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"
 else
-	GOFLAGS=-mod=vendor ginkgo -keepGoing --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial "$SUITE"/parallel
+	GOFLAGS=-mod=vendor ginkgo -keepGoing --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial
```
edcdavid, line 29: Is this a leftover from testing ? parallel suite seems disabled

test/conformance/serial/ptp.go
```diff
@@ -533,6 +534,83 @@ var _ = Describe("["+strings.ToLower(DesiredMode.String())+"-serial]", Serial, f
...
+				pluginConfigExists := false
+				pluginOpts := ""
+				masterConfigs, slaveConfigs := ptphelper.DiscoveryPTPConfiguration(pkg.PtpLinuxDaemonNamespace)
+				ptpConfigs := append(masterConfigs, slaveConfigs...)
+				for _, config := range ptpConfigs {
+					for _, profile := range config.Spec.Profile {
+						if profile.Plugins != nil {
+							for name, opts := range profile.Plugins {
+								if name == "reference" {
+									optsByteArray, _ := json.Marshal(opts)
```
edcdavid, line 582: add some error handling or at least print error

test/pkg/ptphelper/ptphelper.go
```diff
@@ -439,6 +443,29 @@ func PtpEventEnabled() bool {
...
 }
 
+func EnablePTPReferencePlugin() error {
+	ptpOperatorConfig, err := client.Client.PtpV1Interface.PtpOperatorConfigs(pkg.PtpLinuxDaemonNamespace).Get(context.Background(), pkg.PtpConfigOperatorName, metav1.GetOptions{})
+	Expect(err).NotTo(HaveOccurred())
+
+	var plugindata apiextensions.JSON
+	plugindata.Raw = []byte("1")
+
+	(*ptpOperatorConfig.Spec.EnabledPlugins)["reference"] = &plugindata
```
edcdavid, line 453: check that *ptpOperatorConfig.Spec.EnabledPlugins is not  nil before accessing

PR #365, opened 2023-07-27
fix Dockerfile for golang-1.20

jzding:
The golang:1.20 image does not have CGO libraries enabled (CGO_ENABLED=1) so the image will be compiled without c libraries and fail to launch pod.

```
$ oc get pods
NAME                            READY   STATUS             RESTARTS      AGE ptp-operator-84c896d568-q725f   0/1     CrashLoopBackOff   6 (22s ago)   6m19s

$ oc logs -f ptp-operator-84c896d568-q725f
ptp-operator: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ptp-operator)
ptp-operator: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ptp-operator)
```


josephdrichard (approved):
/lgtm

PR #366, opened 2023-08-10
disable Plugin tests

nishant-parekh:
disabling test that is getting stuck


edcdavid (commented):
I would also disable "It("Should check whether test plugin executes", func() {" since these test are related. Also run the test locally to make sure it passes this step

josephdrichard:
Is there a bug open related to this?  What exactly is the issue?

edcdavid:
> Is there a bug open related to this? What exactly is the issue?

not sure, seems to be crashing in CI.

edcdavid (approved):

PR #369, opened 2023-08-31
ptp opeator update docker file to rhel 9

aneeshkp:


jzding (approved):
/lgtm

nishant-parekh:
please do not merge this change as we need to update CI config / ART config before we make this change.. and right now there is no need to move the operator base image to rhel-9

PR #390, opened 2023-10-24
OCPBUGS-22325:  ptp4l process restarted unexpected with dual NIC boundary clocks configured

aneeshkp:
Fix dual nic configuration sometimes triggers restart of process due to non deterministic ordering of the profile name in maps


jzding (approved):
/lgtm

PR #394, opened 2023-10-30
Bug OCPBUGS-22653: fix SNO lease duration

jzding:
leaseDurationSeconds should be 270 seconds for SNO instead of default 15 seconds.

Add back namespace to be consistent with previous releases.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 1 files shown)


aneeshkp:
please add ci to verify  the lease object

aneeshkp:
add lease object if exist validating the duration 
https://github.com/openshift/ptp-operator/blob/master/test/validation/tests/validation.go#L58

jzding:
> add lease object if exist validating the duration https://github.com/openshift/ptp-operator/blob/master/test/validation/tests/validation.go#L58

unit test added

aneeshkp (approved):

PR #415, opened 2024-01-07
OCPBUGS-25470: Fix CI test failures

nishant-parekh:
Match ginkgov2 version used for tests to the vendored version
Update golang version used to build tests

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -4,8 +4,7 @@ set -x
 wget https://go.dev/dl/go1.20.4.linux-amd64.tar.gz
 rm -rf /usr/local/go
 tar -C /usr/local -xzf go1.20.4.linux-amd64.tar.gz
-export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin
-PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin
+export PATH=$PATH:/usr/local/go/bin
 
 which ginkgo
 if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


hack/run-functests.sh
```diff
@@ -22,9 +28,10 @@ export PATH=$PATH:$GOPATH/bin
...
 
+go version
 if [[ $SUITE == *"$VALIDATION_SUIT_SUBSTR"* ]]
 then
 	GOFLAGS=-mod=vendor ginkgo --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"
 else
-	GOFLAGS=-mod=vendor ginkgo -keepGoing --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial "$SUITE"/parallel
-fi
\ No newline at end of file
+	GOFLAGS=-mod=vendor ginkgo -keep-going --output-dir=$JUNIT_OUTPUT_DIR --junit-report=$JUNIT_OUTPUT_FILE -v -p "$SUITE"/serial "$SUITE"/parallel
```
edcdavid, line 36: -keep-going -> --keep-going

hack/run-functests.sh
```diff
@@ -1,15 +1,21 @@
 #!/bin/bash
 set -x
-which ginkgo
+# Install Go 1.20.4
+wget https://go.dev/dl/go1.20.4.linux-amd64.tar.gz
+rm -rf /usr/local/go
+tar -C /usr/local -xzf go1.20.4.linux-amd64.tar.gz
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/go/bin
```
edcdavid, line 7: Why 2 PATH lines? why not just: 
`export PATH=$PATH:/usr/local/go/bin`

edcdavid (approved):
Looks good to me, just a few comments

hack/run-functests.sh
```diff
@@ -1,15 +1,21 @@
 #!/bin/bash
 set -x
-which ginkgo
+# Install Go 1.20.4
+wget https://go.dev/dl/go1.20.4.linux-amd64.tar.gz
+rm -rf /usr/local/go
```
josephdrichard, line 5: So this will remove existing go binary on a user's system?

nishant-parekh: The version of ginkgo used below requires at least this version of golang. This change will enable the CI setup to run the tests as the image used for CI has an older version of golang.. there is an ongoing effort to change the image used after which this change can be rolled back

hack/run-functests.sh
```diff
@@ -1,15 +1,20 @@
...
```
jzding, line 5: If this is a temporary workaround please add a `TODO` comment. For a permanent solution, I suggest 1. check the existing go version and install newer version only when needed. 2. install new go in {ptp-operator repo}/bin and add this in front of the PATH.

jzding (approved):
/lgtm

hack/run-functests.sh
nishant-parekh: I will have a subsequent update to address your comment. Trying to enable CI pipeline again with this change.

PR #416, opened 2024-01-25
OCPBUGS-27738: Add workload annotation and increase CI timeout

jzding:
also increase prometheus_rate_time_window in CI


test/conformance/config/ptptestconfig.yaml
```diff
@@ -25,7 +25,7 @@ soaktest:
       duration: 5
       failure_threshold: 3
       custom_params:
-        prometheus_rate_time_window: "60s"
+        prometheus_rate_time_window: "120s"
```
josephdrichard, line 28: Is this related?

PR #436, opened 2024-02-12
Increase PrometheusQueryRetries from 30 to 60

josephdrichard:
Still getting timeouts on CI at 30 seconds.  Increase to 60.


aneeshkp (approved):
/lgmt

PR #445, opened 2024-02-16
OCPBUGS-29634: fix ci issues

jzding:
- Add pod-security labels to consumer sidecar test namespace.
- Add SetLogger to avoid exception during logging.
- Add prometheus trial poll and wait for 1-10 minutes for the metrics to be ready if the poll fail.
- Adjust prometheus rate time window in relation to cadvisor scrape interval to make sure minimal samples can be scraped.

### DIFF
--- a/hack/run-functests.sh
+++ b/hack/run-functests.sh
@@ -1,10 +1,30 @@
 #!/bin/bash
 set -x
-# Install Go 1.20.4
-wget https://go.dev/dl/go1.20.4.linux-amd64.tar.gz
-rm -rf /usr/local/go
-tar -C /usr/local -xzf go1.20.4.linux-amd64.tar.gz
-export PATH=$PATH:/usr/local/go/bin
+
+# make sure the test runs with specific go vervsion.
+# install go in <ptp-operator-repo>/bin if not already installed
+GO_VERSION=1.20.4
+REPO_BIN_PATH=$(pwd)/bin
+echo "REPO_BIN_PATH is ${REPO_BIN_PATH}"
+export PATH="${REPO_BIN_PATH}/go/bin:$PATH"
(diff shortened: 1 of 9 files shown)


test/pkg/ptptesthelper/ptptesthelper.go
```diff
@@ -374,68 +374,6 @@ func GetPtpPodsPerNode() (map[string][]*corev1.Pod, error) {
 	return podsPerNode, nil
 }
 
-// GetPodsTotalCpuUsage uses prometheus metric "container_cpu_usage_seconds_total"
-// to return the total cpu usage by all the given pods.
-// As each query needs to be done inside one of the prometheus pods, an optional
-// param prometheusPod can be set for that purpose. If it's nil, the function
-// will try to get it on every call.
-func GetPodsTotalCpuUsage(pods []*corev1.Pod, prometheusPod *corev1.Pod) (float64, error) {
```
aneeshkp, line 382: Why this was deleted?

jzding: it is not used anywhere

hack/run-functests.sh
```diff
@@ -1,10 +1,30 @@
 #!/bin/bash
 set -x
-# Install Go 1.20.4
-wget https://go.dev/dl/go1.20.4.linux-amd64.tar.gz
-rm -rf /usr/local/go
-tar -C /usr/local -xzf go1.20.4.linux-amd64.tar.gz
-export PATH=$PATH:/usr/local/go/bin
+
```
nishant-parekh, line 3: This part will be reworked as we have a new CI image that has different versions of golang installed. Discussed with @jzding and we will merge this change before adding the updates for the new image

nishant-parekh (approved):
/lgtm
/approve

PR #449, opened 2024-02-26
OCPBUGS-29634: fix CPU Utilization soak test

jzding:
Remove trial prometheus queries.

Check container cpu usage only if a threshold target for this container is set in the test config. This reduces the number of prometheus queries significantly and avoid queries for kube-rbac-proxy and pause containers.

When retry prometheus queries, increase the rateTimeWindow by 10 seconds with each retry to avoid returning empty data due to not enough samples being captured within the window.


nishant-parekh (approved):
/approve
/lgtm

PR #452, opened 2024-02-29
OCPBUGS-29472:  PTP service annotated with certificate error

aneeshkp:
add new secret for ptp publisher service


nishant-parekh (approved):

nishant-parekh:
we do need to revisit this issue once we are looking to add https support

PR #456, opened 2024-03-11
include haprofiles under PTPSettings

aneeshkp:
Enabling High Availability (HA) for phc2sys in ptpConfig
This PR adds haprofile key to ptpConfig to enable High Availability (HA) for phc2sys .
It achieves this by adding profiles of ptp4l-enabled configurations under the haProfiles key and passing the socket path of those profiles to phc2sys .


jzding (approved):
/lgtm

PR #460, opened 2024-03-19
OCPBUGS-30648: PTP CI TC Failure: PTP socket sharing between pods Run pmc in a new pod on the slave node Should be able to sync using a uds

nishant-parekh:
updated the test assertions to:
   * not assert the contents of /host/secrets directory as it is not affected by the pmc commands in the test stated


jzding (approved):
/lgtm

PR #463, opened 2024-04-11
CNF-11962: PTP CI TC timedout: Event based tests PTP Slave Clock Sync

nishant-parekh:
add missing argument to consumer container to process newly added publisher health checks


jzding (approved):
/lgtm

PR #466, opened 2024-04-24
OCPBUGS-32442: Need a ptpOperatorConfig webhook to prevent users from creating a new event config

aneeshkp:
Added validation webhook , to validate PTPOperatorConfig is always Default

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -40,6 +40,7 @@ webhooks:
     apiVersions:
     - v1
     operations:
+    - CREATE
     - UPDATE
     resources:
     - ptpoperatorconfigs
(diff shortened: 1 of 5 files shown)


config/webhook/manifests.yaml
```diff
@@ -40,6 +40,7 @@ webhooks:
     apiVersions:
     - v1
     operations:
+    - CREATE
```
josephdrichard, line 43: With failurePolicy: Fail, if the webhook service isn't running at initial deployment, this will prevent enabling events?

aneeshkp: By default we don't enable events. 15 secs should be good enuff  to check for its availability

aneeshkp: once default config is created only then user can modify or apply changes to CR to enable events

PR #469, opened 2024-05-09
OCPBUGS-33395: Cannot create PTP BC HA configuration if profile has a dash in a name

aneeshkp:
Update profile name validation  to allow hypen and underscore 
regexp.MustCompile(`^([\w\-_]+)(,\s*([\w-_]+))`)


api/v1/ptpconfig_webhook.go
```diff
@@ -40,7 +40,7 @@ const (
 
 // log is for logging in this package.
 var ptpconfiglog = logf.Log.WithName("ptpconfig-resource")
-var profileRegEx = regexp.MustCompile(`^(\w+)(,\s*([\w-_]+))`)
+var profileRegEx = regexp.MustCompile(`^([\w\-_]+)(,\s*([\w-_]+))`)
```
jzding, line 43: This is wrong, should be `^([\w\-_]+)(,\s*([\w\-_]+))`. Note the second `-` should have an escape `\`.

aneeshkp: In Go, the underscore character _ does not require escaping when used in regular expressions.

jzding: I was referring to the dash, not underscore

jzding (approved):
/lgtm

PR #488, opened 2024-06-21
NO-ISSUE: makefile: fix bundle-build, catalog-build, image target

leo8a:
Fixes some targets in Makefile to comply using the existing CONTAINER_TOOL var.

/cc @aneeshkp


aneeshkp (approved):
/lgtm

PR #497, opened 2024-07-04
CNF-13531: Add missing field description to ptpoperatorconfig API

leo8a:
This PR revisits the PtpOperatorConfig spec and adds missing field description in API, as well as blocks the creation of extra resources of this type, to avoid confusion among our users.

/cc @aneeshkp @jzding @josephdrichard


josephdrichard (approved):
/lgtm
/approved

PR #523, opened 2024-08-26
CI Failure fix for Go Version issue

nishant-parekh:


hack/run-functests.sh
```diff
@@ -38,7 +34,9 @@ else
     rm -rf "$temp_dir"
   fi
 fi
-
+source $HOME/golang-1.22.4
```
jzding, line 37: don't need this. this file does not exist in local run.

jzding (approved):
/lgtm

PR #542, opened 2024-10-07
OCPBUGS-42776: Add webhook validation for event apiVersion

jzding:


api/v1/ptpoperatorconfig_webhook.go
```diff
@@ -45,6 +46,13 @@ func (r *PtpOperatorConfig) validate() error {
 	if r.GetName() != "default" {
 		return errors.New("PtpOperatorConfig name must be 'default'. Only one 'default' PtpOperatorConfig configuration is allowed")
 	}
+
+	if r.Spec.EventConfig != nil && r.Spec.EventConfig.EnableEventPublisher {
+		if r.Spec.EventConfig.ApiVersion != "" && !isValidVersion(r.Spec.EventConfig.ApiVersion) {
```
aneeshkp, line 51: do you need apiVersion nil check?

jzding: it's a string so can't be nil

aneeshkp (approved):

PR #548, opened 2024-10-28
CNF-14533: Event Based Additional Tests

nishant-parekh:
Covered in the PR:
4 additional tests added per the test plan here:
* Loss of GNSS
* Holdover verification.
* Event based tests v1 and v2.

review fixes:
   * update ubxtool command to use COLDBOOT for GNSS lost testing
   * use linuxptp enums for consistency between tests and code.
   * comment out gpsd and gpspipe checks pending bugfix.
   * check only GM clock state for now.
  
TODO: Event verification for v1 and v2 is pending, working on follow up PR after this.


aneeshkp (approved):
/lgtm

PR #549, opened 2024-11-04
Add support for fips

aneeshkp:
Enable  FIPS for PTP  Operator
1.Add features.operators.openshift.io/fips-compliant: "true" annotation in the ClusterServiceVersion manifest of your operator


jzding (approved):
/lgtm

PR #564, opened 2025-01-17
PTP CI TC failure: PTP e2e tests prometheus Metrics reported by PTP pods Should all be reported by prometheus (Intermittent)

aneeshkp:
Limits excessive output: Restricts error reporting to 5 missing metrics (configurable with maxErrors).
Prevents Gomega truncation: Instead of dumping all missing metrics, it provides  summary.
Logs failed queries for debugging instead of stopping execution
Also  handle nil result


jzding (approved):
/lgtm

PR #571, opened 2025-02-10
Upstream to downstream staging main

josephdrichard:


Makefile
```diff
@@ -43,7 +43,7 @@ BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION).0 $(BUNDLE_METADATA_OPTS
 OPERATOR_SDK_VERSION ?= v1.22.0-ocp
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/openshift/origin-ptp-operator:$(VERSION)
```
aneeshkp, line 46: This should still maintain $(version) since this will bemodified directly dowstream and never pull from upstream

bundle/manifests/ptp-operator.clusterserviceversion.yaml
```diff
@@ -465,7 +465,7 @@ spec:
   - ptp
   links:
   - name: Documentation operator
-    url: https://github.com/openshift/ptp-operator
+    url: https://github.com/k8snetworkplumbingwg/ptp-operator
   - name: Documentation operand
     url: https://github.com/openshift/linuxptp-daemon
```
aneeshkp, line 470: operand path should change too ?

config/manifests/bases/ptp-operator.clusterserviceversion.yaml
```diff
@@ -166,7 +166,7 @@ spec:
   - ptp
   links:
   - name: Documentation operator
-    url: https://github.com/openshift/ptp-operator
+    url: https://github.com/k8snetworkplumbingwg/ptp-operator
   - name: Documentation operand
     url: https://github.com/openshift/linuxptp-daemon
```
aneeshkp, line 171: operand path should be upstream

PROJECT
```diff
@@ -34,7 +34,7 @@ resources:
   domain: openshift.io
   group: ptp
   kind: PtpOperatorConfig
-  path: github.com/openshift/ptp-operator/api/v1
+  path: github.com/k8snetworkplumbingwg/ptp-operator/api/v1
```
nocturnalastro, line 37: Should these be replaced?

nocturnalastro: Would it be worth writing a script to replace these using a regex or something like that?

josephdrichard: We want to keep downstream using the same thing as upstream here, so I don't think any reason for that.

jzding (approved):
/lgtm

PR #574, opened 2025-02-11
TEST and test conflict on case-insensitive file systems

rdavid:
TEST and test conflict on case-insensitive file systems, such as the default in macOS.


josephdrichard:
Opened https://github.com/openshift/ptp-operator/pull/575 which is merging this from upstream to downstream to fix this, now that we are upstream first
/close this one

rdavid:
@josephdrichard I was first! :-)

PR #582, opened 2025-03-12
[release-4.12] OCPBUGS-53023: fix compile error in CI

jzding:
The root cause is a compatibility issue between the expected GLIBC version (2.34) and the available version (2.28) in the builder image.
Functions like res_hnok and res_dnok, related to DNS name validation in libresolv.so, were in GLIBC 2.34 (RHEL 9) but not in 2.28 (RHEL 8). GLIBC versions in RHEL can be found here https://access.redhat.com/solutions/38634

Dockerfile.rhel7 is used by ART build. This issue seems only occur in CI environment so we keep Dockerfile.rhel7 intact.


nocturnalastro (approved):
/lgtm

PR #585, opened 2025-03-18
Upstream to downstream staging main

josephdrichard:


rdavid (approved):

test/pkg/event/event.go
```diff
@@ -76,6 +76,17 @@ func GetDefaultApiVersion() string {
 	return "2.0"
 }
 
+// GetConsumerImage ...
+func GetConsumerImage() string {
```
aneeshkp, line 80: Why ccloud-event-proxy consumer code is in ptp operator

jzding (approved):
/lgtm

